PR #34, opened 2024-03-08
Fix output formatting for free

claudex:
The GNU free is using 11 chars wide for number, not 12. This was also causing a misalignment of the numbers with the title.

The source for the 11 chars wide: https://gitlab.com/procps-ng/procps/-/blob/master/src/free.c?ref_type=heads#L336

### DIFF
--- a/src/uu/free/src/free.rs
+++ b/src/uu/free/src/free.rs
@@ -82,9 +82,9 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
             let used = mem_info.total - mem_info.free;
 
             if wide {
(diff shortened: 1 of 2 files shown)


sylvestre:
please don't look at the source of the GNU implementation. it is GPL code and incompatible with this license.

sylvestre:
also, could you please add a test https://github.com/uutils/procps/blob/main/tests/by-util/test_free.rs to make sure we don't regress?
thanks

claudex:
I've added the tests, tell me if you think I should test more.

claudex:
Sorry, I forgot to run `cargo fmt`

sylvestre:
thanks

Issue #21, opened 2024-03-02
free: implement `--line` 


sylvestre:
```
       -L, --line
              Show output on a single line, often used with the -s option to show memory  statis-
              tics repeatedly.
```

eduardorittner:
Wasn't this option implemented in https://github.com/uutils/procps/commit/1f0ae84ddab0b134d5b7b09011b79f3f8e189613 already?

Issue #18, opened 2024-03-02
free: implement the different size display


sylvestre:
```
      -b, --bytes
              Display the amount of memory in bytes.

       -k, --kibi
              Display the amount of memory in kibibytes.  This is the default.

       -m, --mebi
              Display the amount of memory in mebibytes.

       -g, --gibi
              Display the amount of memory in gibibytes.

       --tebi Display the amount of memory in tebibytes.

       --pebi Display the amount of memory in pebibytes.

       --kilo Display the amount of memory in kilobytes. Implies --si.

       --mega Display the amount of memory in megabytes. Implies --si.

       --giga Display the amount of memory in gigabytes. Implies --si.

       --tera Display the amount of memory in terabytes. Implies --si.

       --peta Display the amount of memory in petabytes. Implies --si.

      -h, --human
              Show all output fields automatically scaled to shortest three digit unit  and  dis-
              play the units of print out.  Following units are used.

                B = bytes
                Ki = kibibyte
                Mi = mebibyte
                Gi = gibibyte
                Ti = tebibyte
                Pi = pebibyte

              If  unit  is  missing,  and  you  have  exbibyte  of  RAM or swap, the number is in
              tebibytes and columns might not be aligned with header.
```

Krysztal112233:
finished at #35

cakebaker:
Closing this ticket as everything seems to be implemented, with the exception of `--human`. I opened a separate [ticket](https://github.com/uutils/procps/issues/73) for it.

PR #35, opened 2024-03-21
free: Implement almost all functions of  the `free` commands

Krysztal112233:
## Aims

This pull request aims to implement almost all function about `free` command.

And also fix wron arguments and flags.

## TODO

- [x] Fix `-h` flag's function
- [x] #18
- [ ] Implement `-h` `--human` flag
  - [x] Automatically scale
  - [ ] Scaled shortest three digit unit
- [ ] #21


sylvestre:
great start but 
please don't implement everything into a single PR

Krysztal112233:
> great start but please don't implement everything into a single PR

Okay, next I'll split it into multiple PR.

So what should this PR do next?

src/uu/free/src/free.rs
```diff
@@ -71,43 +78,77 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
...
     match parse_meminfo() {
         Ok(mem_info) => {
-            let buff_cache = if wide {
-                mem_info.buffers
-            } else {
-                mem_info.buffers + mem_info.cached
+            let buff_cache = match wide {
+                true => mem_info.buffers,
+                false => mem_info.buffers + mem_info.cached,
+            };
```
sylvestre, line 88:
```suggestion
            let buff_cache = if wide { mem_info.buffers } else { mem_info.buffers + mem_info.cached };
```

src/uu/free/src/free.rs
```diff
@@ -71,43 +78,77 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
...
-                    mem_info.total,
-                    used,
-                    mem_info.free,
-                    mem_info.shared,
-                    buff_cache + mem_info.reclaimable,
-                    mem_info.available
-                );
+            let used = mem_info.total - mem_info.available;
+
+            match wide {
```
sylvestre, line 95: same, please use an if

src/uu/free/src/free.rs
```diff
@@ -71,43 +78,77 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
...
+                true => {
+                    wide_header();
+                    match human {
+                        true => println!(
```
sylvestre, line 99: same, if

src/uu/free/src/units.rs
```diff
@@ -0,0 +1,139 @@
+use clap::ArgMatches;
+use std::fmt::{self, Display, Formatter};
+
+#[derive(Debug, PartialEq)]
+pub(crate) enum UnitMultiplier {
```
sylvestre, line 5: there isn't a crate for this already?

sylvestre: like https://crates.io/crates/human_bytes

Krysztal112233: Oh, thanks for the reminder.

Krysztal112233:
I have a question about using if instead of match:

Sometimes match looks cleaner and more direct, so why not use match? Because match won't be optimized by branch prediction?

sylvestre:
> Sometimes match looks cleaner and more direct, so why not use match? 
Because match is used when we have several diverse values. 
For such operations, the de facto standard is to use if instead.
 it is also for consistency with coreutils

> Because match won't be optimized by branch prediction?
I hope that, between Rust and LLVM, it will generate the same code.

Krysztal112233:
> https://godbolt.org/z/Ec3aedhWG confirms that is generate the same code

I got it

src/uu/free/src/units.rs
sylvestre: did you look at it ? :)
thanks

Krysztal112233:
completes almost all the most often flags for the `free` command now

src/uu/free/src/units.rs
Krysztal112233: hum I don't think this crate is quite the way it needs to be, because it only supports one of SI and IEC. I used [bytesize](https://crates.io/crates/bytesize)

sylvestre:
Sorry, i only realize thi now. Could you please add tests in 
https://github.com/uutils/procps/blob/main/tests/by-util/test_free.rs
thanks

PR #41, opened 2024-03-27
w: Format time and get cmdline

fortifiedhill:
This change allows w to display time in a similar way to the original procps w. Additionally, w now fetches the cmdline for each entry.

Later, the fetch_cmdline and possibly format_time functions could be seperated out into a shared library for future applications, such as `ps` or `top`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,6 +47,7 @@ xattr = "1.3.1"
 tempfile = "3.9.0"
 rand = { version = "0.8", features = ["small_rng"] }
 bytesize = "1.3.0"
+time = { version = "0.3", features = ["formatting"] }
 
 [workspace.'cfg(any(target_os = "linux", target_os = "android"))'.dependencies]
 utmpx = "0.1"
@@ -75,6 +76,7 @@ tempfile = { workspace = true }
 libc = { workspace = true }
 rand = { workspace = true }
 uucore = { workspace = true, features = ["entries", "process", "signals"] }
+time = { workspace = true, features = ["formatting"] }
 
 [target.'cfg(unix)'.dev-dependencies]
 xattr = { workspace = true }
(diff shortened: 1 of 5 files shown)


sylvestre:
nice, could you please add a test? thanks

fortifiedhill:
> nice, could you please add a test? thanks

No problem, I've added a unit test for each new function. I did this instead of testing the actual output, because it is bound to change in the future and the unit tests may be helpful if/when these function are moved into a different location.

sylvestre:
Sure and thanks  but could you please add integration test in tests/ too? Thanks

fortifiedhill:
> Sure and thanks but could you please add integration test in tests/ too? Thanks

I went ahead and removed the unit tests and added integration tests into tests/ instead. It might be a good idea to treat them as public functions from the start and then move them into a library or separate file, when it is time.

sylvestre:
Please keep the function tests in the same file.
Add tests on the binary like
test_no_header

fortifiedhill:
> Please keep the function tests in the same file. Add tests on the binary like test_no_header

I've put the unit tests where they should be and I've added an integration test in tests/ that checks the current output of w.

Cargo.toml
```diff
@@ -47,6 +47,7 @@ xattr = "1.3.1"
 tempfile = "3.9.0"
 rand = { version = "0.8", features = ["small_rng"] }
 bytesize = "1.3.0"
+time = { version = "0.3", features = ["formatting"] }
```
sylvestre, line 50: would you mind using chrono ? 
we are trying to use only crate for time mgmt

sorry

fortifiedhill: Sure, I've switched to chrono. Also, there is no point in checking the formatting of the cmdline, as it can be empty, a path, or just an executable name.

PR #48, opened 2024-04-01
watch: implement interval

maxer137:
This merge aims to add an interval parsing implementation for watch.
```
      -n, --interval seconds
              Specify update interval.  The command will not allow quicker than 0.1 second inter-
              val,  in  which the smaller values are converted. Both '.' and ',' work for any lo-
              cales. The WATCH_INTERVAL environment can be used to persistently set a non-default
              interval (following the same rules and formatting).
```

In watch.rs, a new function named `parse_interval` is added, which takes in a `&str` and returns a result with `Duration` or an error with `ParseIntError`.

The function will first see if the string contains either a '.' or a ',' character. If it does not, it attempts to parse it into seconds and return a duration with that many seconds.
Otherwise, it will attempt to split the string into the seconds and the remainder.
The remainder is then parsed and zero padded. This way we ensure 0.5 becomes 500000000 nanoseconds, and not 5.
If the string has a missing component (such as `.5` or `5.`) the function will assume 0.
Nanosecond strings that except 9 characters will first be checked if they are valid, then trimmed.
Finally, return the calculated duration or 100 milliseconds. Whichever is longer. This ensures the minimum interval of 0.1 seconds. 

Other thoughts:
1. Currently, I have used `"a".parse::<u8>()?` to get rust to return a `ParseIntError`. This is obviously not ideal, but it seems ParseIntError is private to rust, and we'd need to make a custom return type instead. This is possible, but it seemed overkill for this scenario.
2. There may be a more efficient calculation for padding the nanoseconds. I used 10u32.pow() but maybe a while loops would be more efficient.
3. The `WATCH_INTERVAL` environment variable has not been implemented yet.

### DIFF
--- a/tests/by-util/test_watch.rs
+++ b/tests/by-util/test_watch.rs
@@ -19,9 +19,7 @@ fn test_invalid_interval() {
 
 #[test]
 fn test_no_interval() {
-    let mut p = new_ucmd!()
-        .arg("true")
-        .run_no_wait();
+    let mut p = new_ucmd!().arg("true").run_no_wait();
     p.make_assertion_with_delay(500).is_alive();
     p.kill()
         .make_assertion()
@@ -33,9 +31,7 @@ fn test_no_interval() {
 #[test]
 fn test_valid_interval() {
     let args = vec!["-n", "1.5", "true"];
-    let mut p = new_ucmd!()
-        .args(&args)
-        .run_no_wait();
+    let mut p = new_ucmd!().args(&args).run_no_wait();
     p.make_assertion_with_delay(500).is_alive();
     p.kill()
         .make_assertion()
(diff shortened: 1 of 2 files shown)


tests/by-util/test_watch.rs
```diff
@@ -10,3 +10,44 @@ use crate::common::util::TestScenario;
 fn test_invalid_arg() {
     new_ucmd!().arg("--definitely-invalid").fails().code_is(1);
 }
+
+#[test]
+fn test_invalid_interval() {
+    let args = vec!["-n", "definitely-not-valid", "true"];
+    new_ucmd!().args(&args).fails();
```
sylvestre, line 17: could you please check stderr too here?
(at least with .contains, doesn't have to be a full match)

maxer137: Added in latest commit

PR #53, opened 2024-04-05
w: Implement PCPU and JCPU

fortifiedhill:
The w manpage defines JCPU as the total CPU time of all processes under a given tty. The PCPU is defined as being the total time being used by the command in the "what" field.

For the JCPU implementation, the program creates a hashmap of all terminal numbers and their total time by walking the pid folders in /proc. I don't think this is terrible performance-wise, as execution time is still slightly faster than the original w, when built in release mode.

Total time for a process is calculated by taking the sum of time scheduled in both user mode and kernel mode and dividing that by the system's clock tick.

In terms of tests, I've added a couple of unit tests and updated the test_output_format to make sure the JCPU and PCPU fields are populated, these still need to be formatted, so the tests don't check for formatting.

### DIFF
--- a/src/uu/w/src/w.rs
+++ b/src/uu/w/src/w.rs
@@ -3,16 +3,16 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 
-#[cfg(not(windows))]
+#[cfg(target_os = "linux")]
 use chrono::{self, Datelike};
 use clap::crate_version;
 use clap::{Arg, ArgAction, Command};
-#[cfg(not(windows))]
+#[cfg(target_os = "linux")]
 use libc::{sysconf, _SC_CLK_TCK};
 use std::process;
-#[cfg(not(windows))]
+#[cfg(target_os = "linux")]
 use std::{collections::HashMap, fs, path::Path};
-#[cfg(not(windows))]
+#[cfg(target_os = "linux")]
 use uucore::utmpx::Utmpx;
 use uucore::{error::UResult, format_usage, help_about, help_usage};
 
--- a/src/uu/w/src/w.rs
+++ b/src/uu/w/src/w.rs
@@ -29,7 +29,7 @@ struct UserInfo {
     command: String,
 }
 
-#[cfg(not(windows))]
+#[cfg(target_os = "linux")]
 fn fetch_terminal_jcpu() -> Result<HashMap<u64, f64>, std::io::Error> {
     // Hashmap of terminal numbers and their respective CPU usages
     let mut pid_hashmap = HashMap::new();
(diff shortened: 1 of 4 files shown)


cakebaker:
Did you see that some tests fail on macOS? It seems like there is no `/proc` on macOS.

fortifiedhill:
> Did you see that some tests fail on macOS? It seems like there is no `/proc` on macOS.

Thanks, I've fixed the issue by only enabling /proc based functions on Linux, as I've only tested them there.

src/uu/w/src/w.rs
```diff
@@ -27,7 +29,69 @@ struct UserInfo {
...
+                            pid_hashmap.insert(terminal_number, fetch_pcpu_time(pid)?);
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    Ok(pid_hashmap)
+}
```
sylvestre, line 63: more rust like:

```suggestion
fn fetch_terminal_jcpu() -> Result<HashMap<u64, f64>, std::io::Error> {
    let pid_dirs = fs::read_dir("/proc")?
        .filter_map(Result::ok)
        .filter(|entry| entry.path().is_dir())
        .filter_map(|entry| entry.path().file_name().and_then(|s| s.to_os_string().into_string().ok()))
        .filter_map(|pid_dir_str| pid_dir_str.parse::<i32>().ok());

    let mut pid_hashmap = HashMap::new();
    for pid in pid_dirs {
        let terminal_number = fetch_terminal_number(pid)?;
        let pcpu_time = fetch_pcpu_time(pid)?;
        *pid_hashmap.entry(terminal_number).or_insert(0.0) += pcpu_time;
    }

    Ok(pid_hashmap)
}

```

src/uu/w/src/w.rs
```diff
@@ -27,7 +29,69 @@ struct UserInfo {
...
+
+#[cfg(target_os = "linux")]
+fn fetch_terminal_number(pid: i32) -> Result<u64, std::io::Error> {
+    let stat_path = Path::new("/proc").join(pid.to_string()).join("stat");
+    // Seperate stat and get terminal number, which is at position 6
```
sylvestre, line 68:
```suggestion
    // Separate stat and get terminal number, which is at position 6
```

src/uu/w/src/w.rs
```diff
@@ -27,7 +29,69 @@ struct UserInfo {
...
+    let f = fs::read_to_string(stat_path)?;
+    let stat: Vec<&str> = f.split_whitespace().collect();
+    let terminal_number: u64 = stat[6].parse().unwrap_or_default();
+    Ok(terminal_number)
```
sylvestre, line 72:
```suggestion
    Ok(stat[6].parse().unwrap_or_default())
```

fortifiedhill:
Thanks for the detailed and thorough code review sylvestre! I've tested the changes and have committed them.

fortifiedhill:
> Could you please add the comments back?
> 
> I think your merge brings back the previous issue, the self?

You're talking about the comments that I had before the code review? And I removed the self, as was done in #54, to resolve the conflict, I think that's correct?

PR #61, opened 2024-04-17
`w`: implement --short option

ant0nkress:
# Description

We check for the parameter `-s | --short`. If it is set we exclude printing of `login time`, `JCPU time` and `PCPU time`.

# Testing

Added a test to make sure only necessary values are printed.


cakebaker:
@ant0nkress ping?

cakebaker:
Thanks for your PR!

PR #62, opened 2024-04-18
ci: use --all-targets with clippy

cakebaker:
This PR adds `--all-targets` when running `clippy` to detect issues in tests, too.


cakebaker:
Changes since last push:

* none, just a rebase to show the `clippy` jobs pass now

PR #70, opened 2024-04-24
implement free features

D3V1LC0D3R:


D3V1LC0D3R:
wait im stupid i forgot to test whether it works for count = 1 🥴

D3V1LC0D3R:
i dont know how important that is but our free doesnt throw an error for count is 0

D3V1LC0D3R:
i just realized that swap and total have no human output, gotta fix that

sylvestre:
please add tests in https://github.com/uutils/procps/blob/main/tests/by-util/test_free.rs
thanks

D3V1LC0D3R:
@sylvestre i added some basic tests but they dont cover combinations such as -th

D3V1LC0D3R:
@sylvestre will this pr get merged?

sylvestre:
probably but please be a bit patient :)

tests/by-util/test_free.rs
```diff
@@ -4,11 +4,15 @@
 // file that was distributed with this source code.
 // spell-checker:ignore (words) symdir somefakedir
 
+use core::str;
+
```
cakebaker, line 8: I think this import is not necessary
```suggestion
```

tests/by-util/test_free.rs
```diff
@@ -27,6 +31,19 @@ fn test_free_wide() {
     assert!(!result.stdout_str().contains("buff/cache"));
 }
 
+#[test]
+fn test_free_total() {
+    let result = new_ucmd!().arg("-t").succeeds();
+    assert_eq!(result.stdout_str().lines().count(), 4);
+    assert!(result.stdout_str().lines().collect::<Vec<&str>>()[3].starts_with("Total:"))
```
cakebaker, line 38: As you only care about what's on the last line, I suggest to use `last()` instead of `collect()`.

tests/by-util/test_free.rs
```diff
@@ -27,6 +31,19 @@ fn test_free_wide() {
...
+}
+
+#[test]
+fn test_free_count() {
+    let result = new_ucmd!().args(&["-c", "2", "-s", "0"]).succeeds();
```
cakebaker, line 43: This test is about `-c`, so I would omit the `-s` as it's not necessary.
```suggestion
    let result = new_ucmd!().args(&["-c", "2"]).succeeds();
```

D3V1LC0D3R:
@sylvestre sry xD i was just a bit bored so i cleaned up the code a bit and implemented --lohi, committing after i tested it a bit :)

tests/by-util/test_free.rs
cakebaker, line 1: I think there are at least two tests missing: what happens if you pass `0` as a value to `-c` or `-s`?

D3V1LC0D3R: if -s is 0 the original throws an error for some reason but doesnt for very small values above 0 (that dont get rounded down to 0). if -c is 0 it also throws an error but if -s is defined and -c is undefined it just runs forever.

cakebaker:
Why do you think `--lohi` doesn't work with the original? And your implementation works?

I have my doubts that your implementation works because when I look at your changes, I see no changes made to `parse_meminfo()`. And the keys `HighTotal`, `HighFree`, `LowTotal`, and `LowFree` from `/proc/meminfo` are never read.

D3V1LC0D3R:
parse_meminfo is called in the match

sylvestre:
According to code coverage, it seems that a bunch of lines aren't covered by tests. could you please extend the tests ? thanks

cakebaker:
@D3V1LC0D3R yes, `parse_meminfo()` is called, though I'm talking about it's implementation. If you look at it, the keys I mentioned above are not handled by this function (nor anywhere else in the program), even though those keys look like they contain the `--lohi` information, hence my skepticism ;-)

I think it is easy to mistake the output of `free --lohi` as incorrect. Here is the output from my machine:
```
               total        used        free      shared  buff/cache   available
Mem:         8024760     3501244      964364      561908     4434544     4523516
Low:         8024760     7060396      964364
High:              0           0           0
```
Obviously, the "High" values must be incorrect, they must be greater or equal to the "Low" values. However, "Low" and "High" refer to different memory regions and as there is no high memory on my machine (and the aforementioned keys are not in `/proc/meminfo`), it is shown as `0`.

D3V1LC0D3R:
@cakebaker i thought --lohi was just info about the highest and lowest the memory has been while running free

D3V1LC0D3R:
so i moved my incorrect `--lohi` to `--minmax` and implemented `--lohi` according to the kernel docs. i dont have a mac so somebody else has to implement that. the sysinfo library doesnt have any info on that as far as i can see.

tests/by-util/test_free.rs
D3V1LC0D3R: i forgot about this; changed it

tests/by-util/test_free.rs
D3V1LC0D3R: i did that so i have to wait less 😂

D3V1LC0D3R:
now free should be complete

D3V1LC0D3R:
@sylvestre @cakebaker are there any problems with this pr?

src/uu/free/src/free.rs
```diff
@@ -43,10 +48,62 @@ struct MemInfo {
...
+    high_used: u64,
+    high_free: u64,
+    commit_limit: u64,
+    committed: u64,
+}
+
+impl MemInfo {
+    fn make_new(&self, o: &MemInfo, min: bool) -> MemInfo {
+        // to use min
+        let clos = |a: u64, b: u64| -> u64 {
```
sylvestre, line 64: what means clos ?

src/uu/free/src/free.rs
```diff
@@ -90,6 +155,15 @@ fn parse_meminfo() -> Result<MemInfo, Box<dyn std::error::Error>> {
         swap_free: sys.free_swap(),
         swap_used: sys.total_swap().saturating_sub(sys.free_swap()),
         reclaimable: 0,
+        // i dont know how to get that kind of info, you probably have to either pay money to get the info or pay some money for some developer docs
```
sylvestre, line 158: this document doesn't provide much value, please make it more actionable

src/uu/free/src/free.rs
```diff
@@ -101,88 +175,198 @@ fn parse_meminfo() -> Result<MemInfo, Box<dyn std::error::Error>> {
     Ok(MemInfo::default())
 }
 
+// print total - used - free combo that is used for everything except memory for now
+// free can be negative if the memory is overcommitted so it has to be signed
+fn tuf_combo<F>(name: &str, total: u64, used: u64, free: i128, f: F)
+where
+    F: Fn(u64) -> String,
+{
+    // imo ugly hack to convert negative values
```
sylvestre, line 184:
```suggestion
    // ugly hack to convert negative values
```

src/uu/free/src/free.rs
D3V1LC0D3R: clos is a closure that can switch between selecting the minimum value or the maximum value

src/uu/free/src/free.rs
D3V1LC0D3R: the only thing i found is https://apple.stackexchange.com/questions/101488/os-xs-free-or-proc-meminfo-equivalent but i dont use macos so i dont actually know where you could get detailed memory information from.

src/uu/free/src/free.rs
sylvestre: ok, please rename it for what is doing

sylvestre: did you try to simplify it this way?

|a: u64, b: u64| if min { a.min(b) } else { a.max(b) };

src/uu/free/src/free.rs
```diff
@@ -101,88 +174,198 @@ fn parse_meminfo() -> Result<MemInfo, Box<dyn std::error::Error>> {
...
+    let mut max: Option<MemInfo> = None;
+
+    while count > 0 || infinite {
+        // prevent underflow
+        if !infinite {
+            count -= 1;
+        }
+        match parse_meminfo() {
+            Ok(mem_info) => {
+                // dont calculate stuff we dont need in non-lohi situations
```
sylvestre, line 226:
```suggestion
                // Don't calculate stuff we don't need in non-lohi situations
```

src/uu/free/src/free.rs
D3V1LC0D3R: i also just realized that MaxMem should have the lowest free value

src/uu/free/src/free.rs
```diff
@@ -43,10 +48,66 @@ struct MemInfo {
...
+        fn compare(a: u64, b: u64, min: bool) -> u64 {
+            if min {
+                a.min(b)
+            } else {
+                a.max(b)
+            }
+        }
+
+        // invert the behaviour for free and avail. memory
+        let clos = |a: u64, b: u64| compare(a, b, min);
```
sylvestre, line 73: please rename clos and anticlos

like
chooseMinMax
and
reverse_chooseMinMax

src/uu/free/src/free.rs
```diff
@@ -70,29 +70,29 @@ impl MemInfo {
...
-            low_total: clos(self.low_total, o.low_total),
-            low_used: clos(self.low_used, o.low_used),
-            low_free: anticlos(self.low_free, o.low_free),
-            high_total: clos(self.high_total, o.high_total),
-            high_used: clos(self.high_used, o.high_used),
-            high_free: anticlos(self.high_free, o.high_free),
-            commit_limit: clos(self.commit_limit, o.commit_limit),
-            committed: clos(self.committed, o.committed),
+            total: choose_min_max(self.total, o.total),
+            free: reverse_choose_min_max(self.free, o.free),
```
sylvestre, line 79: why not 

```suggestion
            total: compare(self.total, o.total, min),
            free: compare(self.free, o.free, !min),
```

src/uu/free/src/free.rs
```diff
@@ -43,10 +48,66 @@ struct MemInfo {
...
+        }
+
+        // invert the behaviour for free and avail. memory
+        let choose_min_max = |a: u64, b: u64| compare(a, b, min);
+        let reverse_choose_min_max = |a: u64, b: u64| compare(a, b, !min);
+
+        // looping over structs only exists in serde
+        MemInfo {
+            total: compare(self.total, o.total, min),
+            free: compare(self.free, o.free, !min),
```
sylvestre, line 79: sorry, i meant for everything :)

D3V1LC0D3R:
@sylvestre anything else?

src/uu/free/src/free.rs
```diff
@@ -101,88 +175,198 @@ fn parse_meminfo() -> Result<MemInfo, Box<dyn std::error::Error>> {
...
 pub fn uumain(args: impl uucore::Args) -> UResult<()> {
     let matches = uu_app().try_get_matches_from(args)?;
-    let wide = matches.get_flag("wide");
 
+    let wide = matches.get_flag("wide");
     let human = matches.get_flag("human");
-
     let si = matches.get_flag("si");
+    let total = matches.get_flag("total");
+    let minmax = matches.get_flag("minmax");
```
sylvestre, line 202: I don't see minmax in my free binary
 free --help|grep min    
doesn't return anything ?!

D3V1LC0D3R: --minmax thats a new feature i implemented

sylvestre: sorry but for now, we want free to be a dropped in replacement

New features should be done in a separate PR and discussed :)

D3V1LC0D3R: so i should remove this feature? ok :(

sylvestre: yes, it needs to be discussed first
thanks

sylvestre:
Code coverage seems to complain that a few lines aren't covered by test, could you please add a few tests? thanks

D3V1LC0D3R:
@sylvestre i think that is out of date. how can i check that locally? cargo test only tests and doesnt provide any coverage info as far as i know.

D3V1LC0D3R:
i found the problem. i forgot to remove the test of --minmax

src/uu/free/src/free.rs
```diff
@@ -43,10 +47,19 @@ struct MemInfo {
...
+    high_total: u64,
+    high_used: u64,
+    high_free: u64,
+    commit_limit: u64,
+    committed: u64,
 }
 
 #[cfg(target_os = "linux")]
 fn parse_meminfo() -> Result<MemInfo, Error> {
+    // kernel docs: https://www.kernel.org/doc/html/latest/filesystems/proc.html
```
cakebaker, line 62:
```suggestion
    // kernel docs: https://www.kernel.org/doc/html/latest/filesystems/proc.html#meminfo
```

cakebaker:
I think there is still some misunderstanding about `--lohi` and thus its implementation is incorrect :|

Here's what I get on my machine:
```
$ cargo run free --lohi
               total        used        free      shared  buff/cache   available
Mem:         8024716     3797944     1526748      588260     3632572     4226772
Low:               0           0           0
High:              0           0           0
Swap:              0           0           0
$ free --lohi
               total        used        free      shared  buff/cache   available
Mem:         8024716     3774532     1550112      585096     3629456     4250184
Low:         8024716     6474604     1550112
High:              0           0           0
Swap:              0           0           0
```
In the code, you read `HighTotal` and the other related keys from `/proc/meminfo`. That's correct, however, it's possible that those keys are not available in `/proc/meminfo`, like on my machine. And this case is currently not handled.

src/uu/free/src/free.rs
```diff
@@ -63,6 +76,14 @@ fn parse_meminfo() -> Result<MemInfo, Error> {
                 "SwapTotal" => mem_info.swap_total = parsed_value,
                 "SwapFree" => mem_info.swap_free = parsed_value,
                 "SReclaimable" => mem_info.reclaimable = parsed_value,
+                "LowTotal" => mem_info.low_total = parsed_value,
+                "LowUsed" => mem_info.low_used = parsed_value,
+                "LowFree" => mem_info.low_free = parsed_value,
+                "HighTotal" => mem_info.high_total = parsed_value,
+                "HighUsed" => mem_info.high_used = parsed_value,
```
cakebaker, line 83: `LowUsed` and `HighUsed` are not keys in `/proc/meminfo`, see https://www.kernel.org/doc/html/latest/filesystems/proc.html#meminfo

D3V1LC0D3R:
@cakebaker it should be fixed now from what i understand

D3V1LC0D3R:
i totally forgot about the macos part

src/uu/free/src/free.rs
```diff
@@ -101,88 +135,157 @@ fn parse_meminfo() -> Result<MemInfo, Box<dyn std::error::Error>> {
...
+                    mem_info.buffers + mem_info.cached
+                };
+                let cache = if wide { mem_info.cached } else { 0 };
+                let used = mem_info.total - mem_info.available;
+
+                // function that converts the number to the correct string
+                let n2s = |x| match human {
+                    true => humanized(x, si),
+                    false => convert(x).to_string(),
+                };
```
cakebaker, line 194: I suggest to use an `if/else` instead of a `match` as it is more idiomatic.

Otherwise, I think the PR is ready to merge.

D3V1LC0D3R: fixed

cakebaker:
Thanks for your PR :)

PR #95, opened 2024-05-29
pgrep: Add command `pgrep`

### LINKED ISSUE #26: Add command `pgrep`
```
       pgrep - look up for processes based on name and other at-
       tributes
```

### LINKED ISSUE #94: Add command `pgrep`
```
       pgrep - look up, signal, or wait for processes based on
       name and other attributes
```

Krysztal112233:
fix #26


src/uu/pgrep/pgrep.md
```diff
@@ -0,0 +1,7 @@
+# pgrep
+
+```
+pgrep [options] <program-name>
```
cakebaker, line 4:
```suggestion
pgrep [options] <pattern>
```

src/uu/pgrep/src/pgrep.rs
```diff
@@ -0,0 +1,126 @@
...
+    let default_delimiter = "\n".into();
+    let delimiter = matches
+        .get_one::<String>("delimiter")
+        .unwrap_or(&default_delimiter);
+
+    let mut programs = matches
+        .get_many::<String>("program-name")
+        .expect("need at least one program-name")
+        .map(|it| it.to_string())
+        .collect::<Vec<_>>();
```
cakebaker, line 28: My suggestion is to use `pattern` instead of `programs` and `program-name` because it's the name used in the man page.

src/uu/pgrep/src/pgrep.rs
```diff
@@ -0,0 +1,126 @@
...
+            // arg!(-x     --exact                 "match exactly with the command name"),
+            // arg!(-F     --pidfile <file>        "read PIDs from file"),
+            // arg!(-L     --logpidfile            "fail if PID file is not locked"),
+            // arg!(-r     --runstates <state>     "match runstates [D,S,Z,...]"),
+            // arg!(       --ns <PID>              "match the processes that belong to the same namespace as <pid>"),
+            // arg!(       --nslist <ns>       ... "list which namespaces will be considered for the --ns option."),
+        ])
+        .arg(
+            Arg::new("program-name")
+                .help("Name of the program to find the PID of")
```
cakebaker, line 123: See my previous comment.

src/uu/pgrep/src/pgrep.rs
```diff
@@ -0,0 +1,126 @@
...
+    let matches = uu_app().try_get_matches_from(args)?;
+
+    let default_delimiter = "\n".into();
+    let delimiter = matches
+        .get_one::<String>("delimiter")
+        .unwrap_or(&default_delimiter);
+
+    let mut programs = matches
+        .get_many::<String>("program-name")
+        .expect("need at least one program-name")
```
cakebaker, line 26: I think this assumption is incorrect, because something like `pgrep --oldest` is a valid command.

Krysztal112233: Oh, thanks for the reminder. fixed.

Krysztal112233:
regex is too slow, need to write algorithm for parsing /proc/self/stat

```
cargo run --release -p uu_pgrep -- -n  17.00s user 0.73s system 107% cpu 16.563 total
```

Krysztal112233:
after rewrite algorithm

```
./pgrep -o  0.04s user 0.05s system 98% cpu 0.092 total
./pgrep -n  0.04s user 0.05s system 98% cpu 0.089 total
```

src/uu/pgrep/src/pgrep.rs
```diff
@@ -0,0 +1,188 @@
...
+const USAGE: &str = help_usage!("pgrep.md");
+
+#[uucore::main]
+pub fn uumain(args: impl uucore::Args) -> UResult<()> {
+    let matches = uu_app().try_get_matches_from(args)?;
+
+    let default_delimiter = "\n".into();
+    let delimiter = matches
+        .get_one::<String>("delimiter")
+        .unwrap_or(&default_delimiter);
```
cakebaker, line 24: The `default_delimiter` variable is unnecessary because you already defined a default delimiter in the `clap` settings.

src/uu/pgrep/src/pgrep.rs
```diff
@@ -0,0 +1,188 @@
...
+
+    let mut programs = matches
+        .get_many::<String>("pattern")
+        .unwrap_or_default()
+        .map(|it| it.to_string())
+        .collect::<Vec<_>>();
```
cakebaker, line 30: I think this is a bit misleading because there can be only one pattern (or none).

src/uu/pgrep/src/pgrep.rs
```diff
@@ -0,0 +1,188 @@
...
+
+    let should_ignore_case = matches.get_flag("ignore-case");
+    let should_inverse = matches.get_flag("inverse");
+
+    let flag_newest = matches.get_flag("newest");
+    let flag_oldest = matches.get_flag("oldest");
+
+    if flag_newest != flag_oldest {
```
cakebaker, line 38: I think the following block works fine for `pgrep -o` and `pgrep -n`. However, there is an additional use case with a pattern that's currently not covered:
```
$ pgrep konsole
1236
1657
1881
1914
1936
$ pgrep -o konsole
1236
$ pgrep -n konsole
1936

$ cargo run pgrep konsole
1236
1657
1881
1914
1936
$ cargo run pgrep -o konsole 
1
$ cargo run pgrep -n konsole 
8237
```

src/uu/pgrep/src/pgrep.rs
Krysztal112233: can you give examples of this usage?

src/uu/pgrep/src/pgrep.rs
Krysztal112233: I'll fix it later

sylvestre:
for benchmarking, you should use hyperfine :)

src/uu/pgrep/src/pgrep.rs
```diff
@@ -0,0 +1,185 @@
...
+
+use clap::{arg, crate_version, Arg, ArgAction, ArgGroup, Command};
+use pid::walk_pid;
+use uucore::{error::UResult, format_usage, help_about, help_usage};
+
+const ABOUT: &str = help_about!("pgrep.md");
+const USAGE: &str = help_usage!("pgrep.md");
+
+#[uucore::main]
+pub fn uumain(args: impl uucore::Args) -> UResult<()> {
```
sylvestre, line 18: this function is a bit long
what about something like 
#[uucore::main]
pub fn uumain(args: impl uucore::Args) -> UResult<()> {
    let matches = uu_app().try_get_matches_from(args)?;
    let delimiter = matches.get_one::<String>("delimiter").unwrap();

    if let Some(result) = handle_oldest_newest(&matches)? {
        println!("{}", result);
        return Ok(());
    }

    let programs = collect_programs(&matches);
    let result = search_processes(&programs, &matches);
    print_results(result, delimiter, &matches);

    Ok(())
} 

?

sylvestre: (something like that)

src/uu/pgrep/src/pgrep.rs
cakebaker: An example:
```
$ pgrep -o
1
$ pgrep -o konsole
1245
$ pgrep -o konsole firefox
pgrep: only one pattern can be provided
```

src/uu/pgrep/src/pgrep.rs
Krysztal112233: It works fine now.

```shell
❯ cargo run -p uu_pgrep -- -o zsh zsh
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.03s
     Running `/home/krysztal/Desktop/project/uutils/procps/target/debug/pgrep -o zsh zsh`
pgrep: only one pattern can be provided
❯ cargo run -p uu_pgrep -- -o zsh
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.03s
     Running `/home/krysztal/Desktop/project/uutils/procps/target/debug/pgrep -o zsh`
6174
```

src/uu/pgrep/src/pgrep.rs
Krysztal112233: fixed 

```shell
❯ cargo run -p uu_pgrep -- -o zsh zsh
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.03s
     Running `/home/krysztal/Desktop/project/uutils/procps/target/debug/pgrep -o zsh zsh`
pgrep: only one pattern can be provided
```

src/uu/pgrep/src/pgrep.rs
```diff
@@ -0,0 +1,216 @@
...
+                name.into()
+            };
+
+            patterns.iter().any(|it| name.contains(it)) ^ should_inverse
+        })
+        .collect::<Vec<_>>()
+}
+
+// Make -o and -n as a group of args
+fn handle_oldest_newest(matches: &ArgMatches, patterns: &[String]) -> Option<()> {
```
cakebaker, line 70: The `Option<()>` return type is a hack, it's a disguised bool ;-)  I think it would make more sense to return a `Option<String>`, with the string being a pid. And doing the printing outside of this function.

src/uu/pgrep/src/pgrep.rs
```diff
@@ -0,0 +1,216 @@
...
+    // None => Go on
+    if handle_oldest_newest(&matches, &patterns).is_some() {
+        return Ok(());
+    }
+
+    handle_normal_pid_collect(&matches, &patterns);
+
+    Ok(())
+}
+fn collect_arg_patterns(matches: &ArgMatches) -> Vec<String> {
```
cakebaker, line 31:
```suggestion
}

fn collect_arg_patterns(matches: &ArgMatches) -> Vec<String> {
```

src/uu/pgrep/src/pgrep.rs
```diff
@@ -0,0 +1,216 @@
...
+fn handle_oldest_newest(matches: &ArgMatches, patterns: &[String]) -> Option<()> {
+    let flag_newest = matches.get_flag("newest");
+    let flag_oldest = matches.get_flag("oldest");
+
+    if flag_newest != flag_oldest {
+        // Only accept one pattern.
+        if !patterns.is_empty() && patterns.len() != 1 {
+            println!("pgrep: only one pattern can be provided");
+            return Some(());
+        }
```
cakebaker, line 79: I think it is a general feature that only one pattern can be provided (for example, `-v` and `-c` show the same error if you provide more than one pattern), and so this check should be in a different place.

src/uu/pgrep/src/pgrep.rs
Krysztal112233: Maybe it's better to evaluate result string inside the function to return an `Option<String>`, and print it outside the function?

cakebaker: Yes, probably.

Krysztal112233:
I came up with a better idea, I think make the `handle_*` method just collect the required pid and let the `uumain` method decide the output format based on the flag, the code will be more understandable and concise than before.

src/uu/pgrep/src/pgrep.rs
```diff
@@ -0,0 +1,278 @@
...
+pub fn uumain(args: impl uucore::Args) -> UResult<()> {
+    let matches = uu_app().try_get_matches_from(args)?;
+    let patterns = collect_arg_patterns(&matches);
+
+    let Some(result) =
+        collect_oldest_newest(&matches, &patterns).or(collect_normal_pid(&matches, &patterns))
+    else {
+        println!("pgrep: no matching criteria specified");
+        println!("Try `pgrep --help' for more information");
+        return Ok(());
```
cakebaker, line 26: This is an error situation and hence should return an error. You can probably use an `USimpleError`.

src/uu/pgrep/src/pgrep.rs
```diff
@@ -0,0 +1,278 @@
...
+        println!("{}", result);
+    };
+
+    Ok(())
+}
+
+fn collect_arg_patterns(matches: &ArgMatches) -> Vec<String> {
+    let should_ignore_case = matches.get_flag("ignore-case");
+
+    let programs = matches
```
cakebaker, line 65: The `programs` variable name is misleading because it represents a collection of patterns. And so my suggestion is to rename it to `patterns`.

src/uu/pgrep/src/pgrep.rs
```diff
@@ -0,0 +1,278 @@
...
+        .map(|it| it.to_string());
+
+    if should_ignore_case {
+        programs.map(|it| it.to_lowercase()).collect::<Vec<_>>()
+    } else {
+        programs.collect()
+    }
+}
+
+fn collect_pid(matches: &ArgMatches, patterns: &[String]) -> Vec<PidEntry> {
```
cakebaker, line 77: This function returns a collection and so my suggestion is to use the plural of "pid" in the function name: `collect_pids`.

src/uu/pgrep/src/pgrep.rs
```diff
@@ -0,0 +1,278 @@
...
+            }
+        } else {
+            true
+        };
+
+        if (name_matched || tty_matched) ^ should_inverse {
+            Some(it)
+        } else {
+            None
+        }
```
cakebaker, line 128: Returning an `Option` is odd. A predicate usually returns a `bool`.

src/uu/pgrep/src/pgrep.rs
```diff
@@ -0,0 +1,278 @@
...
+    };
+
+    let mut result = Vec::new();
+
+    for ele in walk_pid() {
+        if let Some(ele) = predicate_closure(ele) {
+            result.push(ele)
+        }
+    }
```
cakebaker, line 137: I don't like the `ele` variable name ;-) As you are dealing with pids, why not name it `pid`? However, I'm not sure whether you even need such a variable. In this block you filter pids based on a predicate and so it might be a candidate for using `filter`.

src/uu/pgrep/src/pgrep.rs
```diff
@@ -0,0 +1,278 @@
...
+
+    result
+}
+
+// Make -o and -n as a group of args
+fn collect_oldest_newest(matches: &ArgMatches, patterns: &[String]) -> Option<Vec<PidEntry>> {
```
cakebaker, line 143: Why do you return an `Option<Vec<PidEntry>>`? Conceptually, there can be at most one newest/oldest pid and so `Option<PidEntry>` should be enough.

src/uu/pgrep/src/pgrep.rs
```diff
@@ -0,0 +1,278 @@
...
+}
+
+fn collect_normal_pid(matches: &ArgMatches, patterns: &[String]) -> Option<Vec<PidEntry>> {
+    let result = collect_pid(matches, patterns);
+
+    // Normal output
+    let result = result.into_iter().collect::<Vec<_>>();
+
+    Some(result)
+}
```
cakebaker, line 231: I don't understand this function. What does "normal" mean in this context? Why is it needed? And why does it return an `Option`, even though it never returns `None`?

cakebaker (commented):
A general feedback: please add more tests :)

src/uu/pgrep/src/pgrep.rs
Krysztal112233: I renamed this it to `evaluate` because `it` needs to be mutable within itself (which have side effects). Using a bool as a return value could cause performance degradation.

src/uu/pgrep/src/pgrep.rs
Krysztal112233: Because the `it` passed in is actually `mut it: PidEntry` so I can't use `filter` X<

src/uu/pgrep/src/pgrep.rs
```diff
@@ -0,0 +1,300 @@
...
+    }
+    // End Pattern Check //
+
+    let Some(result) = handle_oldest_newest(&matches).or(handle_matching_pids(&matches)) else {
+        return Err(USimpleError::new(
+            2,
+            "no matching criteria specified\nTry `pgrep --help' for more information",
+        ));
+    };
+
```
Krysztal112233, line 51: Referenced from pgrep's manpage, there are might be no processes matched

> EXIT STATUS
> 
> 0
>     One or more processes matched the criteria. For pkill and pidwait, one or more processes must also have been successfully signalled or waited for.
> 1
>     No processes matched or none of them could be signalled.
> 2
>     Syntax error in the command line.
> 3
>     Fatal error: out of memory etc.

How should I do this so that the program exit status is 1?

src/uu/pgrep/src/pgrep.rs
Krysztal112233: I want `handle_oldest_newest` and `handle_matching_pids` to have the same return type because they are mutually exclusive and one and only one of the functions will execute properly. This makes it easy to unify the handling of detail flags like `-l` `-a` and so on.

src/uu/pgrep/src/pgrep.rs
cakebaker: I'm not sure I understand your question correctly: there is a `set_exit_code` function in `uucore`.

Cargo.toml
```diff
@@ -42,9 +42,9 @@ xattr = "1.3.1"
 tempfile = "3.10.1"
 rand = { version = "0.8.5", features = ["small_rng"] }
 bytesize = "1.3.0"
-chrono = { version = "0.4.38", default-features = false, features = [
-  "clock",
-] }
+chrono = { version = "0.4.38", default-features = false, features = ["clock"] }
+walkdir = "2.5.0"
+once_cell = "1.19.0"
```
cakebaker, line 47: I think this dependency is unnecessary. Some of its functionality has been added to the standard library in Rust 1.70, including the [`get_or_init`](https://doc.rust-lang.org/std/cell/struct.OnceCell.html#method.get_or_init) function you use.

src/uu/pgrep/src/pgrep.rs
Krysztal112233: Yes that't the things I need, thanks :)

src/uu/pgrep/src/pgrep.rs
cakebaker: Hm, to me it looks like you are solving something at the code level that could be solved at the conceptual level.

Forget the code for a moment and think about what (a part of) `pgrep` on a conceptual level does:

* it creates a list of PIDs,
* sorts the list,
* filters the list,
* and uses the list to generate some output

"some output" can be whatever: all elements of the list, the first element, the last element, the number of elements.

Now, back to the code. Based on their name, I would put `handle_matching_pids` on the third step of this simple model whereas `handle_oldest_newest` is on the fourth step. And so there is a conflict on the conceptual level if you look at those functions as mutually exclusive ;-)

Of course, it's possible that this model is incorrect/incomplete, but I think it provides at least some food for thought.

Krysztal112233: Oh, that's exactly what I was thinking! If I change the `handle_*` function to be a filter, then the whole code will be easier to understand

src/uu/pgrep/src/pgrep.rs
```diff
@@ -0,0 +1,299 @@
...
+    // Verifying regex pattern
+    // And put it into static `REGEX`
+    if matches.get_flag("full") {
+        let regex = Regex::new(pattern.first().unwrap())
+            .map_err(|e| USimpleError::new(1, e.to_string()))?;
+        REGEX.set(regex).unwrap();
+    }
+
+    //// Collect pids ////
```
sylvestre, line 54:
```suggestion
    // Collect pids
```

src/uu/pgrep/src/pgrep.rs
```diff
@@ -0,0 +1,299 @@
...
+    let pids = collect_matched_pids(&matches);
+
+    //// Filtering pids ////
```
sylvestre, line 57: same comment about the //

src/uu/pgrep/src/pgrep.rs
```diff
@@ -0,0 +1,312 @@
...
+
+    let mut entry = if flag_newest {
+        pids.first()
+    } else {
+        pids.last()
+    }
+    .cloned()
+    .unwrap();
+
+    let sort = |start_time: u64| {
```
sylvestre, line 239: it would be nice to add some comments here

src/uu/pgrep/src/process.rs
```diff
@@ -0,0 +1,446 @@
...
+    inner_stat: String,
+
+    cached_status: Option<Rc<HashMap<String, String>>>,
+    cached_stat: Option<Rc<Vec<String>>>,
+
+    cached_start_time: Option<u64>,
+    cached_tty: Option<Rc<HashSet<TerminalType>>>,
+}
+
+impl PidEntry {
```
sylvestre, line 172: same, some comments would be appreciated

sylvestre:
codecov shows that some stuff could be tested :)
see:
https://github.com/uutils/procps/pull/95/files

Krysztal112233:
The pgrep command is implemented as far as I can understand it, can I see what else needs to be improved?

tests/by-util/test_pgrep.rs
```diff
@@ -0,0 +1,83 @@
...
+        .code_is(1);
+}
+
+#[cfg(target_os = "linux")]
+#[test]
+fn test_delimiter() {
+    let binding = new_ucmd!().arg("sh").arg("-d |").succeeds();
+    let output = binding.code_is(0).stdout_str();
+
+    assert!(output.contains('|'))
```
cakebaker, line 66: You can simplify this code by chaining the different function calls: `.succeeds().code_is(0).stdout_contains("|");`

tests/by-util/test_pgrep.rs
```diff
@@ -0,0 +1,83 @@
...
+}
+
+#[test]
+#[cfg(target_os = "linux")]
+fn test_valid_regex() {
+    new_ucmd!()
+        .arg("NO_PROGRAM*")
+        .arg("--exact")
+        .fails()
+        .code_is(1);
```
cakebaker, line 57: While this test works fine, it fails when using a regex that matches, for example `a*`.

tests/by-util/test_pgrep.rs
```diff
@@ -0,0 +1,83 @@
...
+#[test]
+#[cfg(target_os = "linux")]
+fn test_full() {
+    new_ucmd!().arg("sh").arg("--full").succeeds().code_is(0);
+}
+
+#[test]
+#[cfg(target_os = "linux")]
+fn test_invalid_regex() {
+    new_ucmd!().arg("{(*").arg("--exact").fails().code_is(2);
```
cakebaker, line 47: While this test works fine, if you omit the `--exact` argument it fails. The behavior should be the same in both cases.

src/uu/pgrep/src/process.rs
```diff
@@ -0,0 +1,468 @@
...
+        );
+
+        assert!(TerminalType::try_from("value").is_err());
+        assert!(TerminalType::try_from("TtyS10").is_err());
+    }
+
+    #[test]
+    fn test_run_state_conversion() {
+        assert_eq!(RunState::try_from("R").unwrap(), RunState::Running);
+        assert_eq!(RunState::try_from("R").unwrap(), RunState::Running);
```
cakebaker, line 404: It's a duplicate of line 403
```suggestion
```

src/uu/pgrep/src/process.rs
```diff
@@ -0,0 +1,468 @@
...
+            .flat_map(|it| it.read_link())
+            .flat_map(TerminalType::try_from)
+            .collect::<HashSet<_>>();
+
+        assert_eq!(pid_entry.ttys().unwrap(), result.into())
+    }
+
+    #[test]
+    fn test_stat_split() {
+        let case="32 (idle_inject/3) S 2 0 0 0 -1 69238848 0 0 0 0 0 0 0 0 -51 0 1 0 34 0 0 18446744073709551615 0 0 0 0 0 0 0 2147483647 0 0 0 0 17 3 50 1 0 0 0 0 0 0 0 0 0 0 0";
```
cakebaker, line 459:
```suggestion
        let case = "32 (idle_inject/3) S 2 0 0 0 -1 69238848 0 0 0 0 0 0 0 0 -51 0 1 0 34 0 0 18446744073709551615 0 0 0 0 0 0 0 2147483647 0 0 0 0 17 3 50 1 0 0 0 0 0 0 0 0 0 0 0";
```

src/uu/pgrep/src/process.rs
```diff
@@ -0,0 +1,468 @@
...
+        assert!(stat_split(case)[1] == "idle_inject/3");
+
+        let case ="3508 (sh) S 3478 3478 3478 0 -1 4194304 67 0 0 0 0 0 0 0 20 0 1 0 11911 2961408 238 18446744073709551615 94340156948480 94340157028757 140736274114368 0 0 0 0 4096 65538 1 0 0 17 8 0 0 0 0 0 94340157054704 94340157059616 94340163108864 140736274122780 140736274122976 140736274122976 140736274124784 0";
```
cakebaker, line 462:
```suggestion
        let case = "3508 (sh) S 3478 3478 3478 0 -1 4194304 67 0 0 0 0 0 0 0 20 0 1 0 11911 2961408 238 18446744073709551615 94340156948480 94340157028757 140736274114368 0 0 0 0 4096 65538 1 0 0 17 8 0 0 0 0 0 94340157054704 94340157059616 94340163108864 140736274122780 140736274122976 140736274122976 140736274124784 0";
```

src/uu/pgrep/src/process.rs
```diff
@@ -0,0 +1,468 @@
...
+        assert!(stat_split(case)[1] == "sh");
+
+        let case="47246 (kworker /10:1-events) I 2 0 0 0 -1 69238880 0 0 0 0 17 29 0 0 20 0 1 0 1396260 0 0 18446744073709551615 0 0 0 0 0 0 0 2147483647 0 0 0 0 17 10 0 0 0 0 0 0 0 0 0 0 0 0 0";
```
cakebaker, line 465:
```suggestion
        let case = "47246 (kworker /10:1-events) I 2 0 0 0 -1 69238880 0 0 0 0 17 29 0 0 20 0 1 0 1396260 0 0 18446744073709551615 0 0 0 0 0 0 0 2147483647 0 0 0 0 17 10 0 0 0 0 0 0 0 0 0 0 0 0 0";
```

src/uu/pgrep/src/process.rs
```diff
@@ -0,0 +1,468 @@
...
+            .map_err(|_| io::ErrorKind::InvalidData)?;
+
+        self.cached_start_time = Some(time);
+
+        Ok(time)
+    }
+
+    /// Fetch start time from [ProcessInformation::cached_stat]
+    ///
+    /// - [The /proc Filesystem: Table 1-4](https://docs.kernel.org/filesystems/proc.html#id10)
```
cakebaker, line 275: This comment describes another function.

src/uu/pgrep/src/process.rs
```diff
@@ -0,0 +1,468 @@
...
+    pub fn run_state(&mut self) -> Result<RunState, io::Error> {
+        RunState::try_from(self.stat()?.get(2).unwrap().as_str())
+    }
+
+    /// This function will scan the `/proc/<pid>/df` directory
```
cakebaker, line 280:
```suggestion
    /// This function will scan the `/proc/<pid>/fd` directory
```

src/uu/pgrep/src/process.rs
```diff
@@ -0,0 +1,468 @@
...
+            .map(|it| (it.0.to_string(), it.1.trim_start().to_string()))
+            .collect::<HashMap<_, _>>();
+
+        let result = Rc::new(result);
+        self.cached_status = Some(Rc::clone(&result));
+        Rc::clone(&result)
+    }
+
+    /// Collect information from `/proc/<pid>/stat` file
+    fn stat(&mut self) -> Result<Rc<Vec<String>>, io::Error> {
```
cakebaker, line 239: Why do you return a `Result`? Seems unnecessary to me.

src/uu/pgrep/src/process.rs
```diff
@@ -0,0 +1,468 @@
...
+    pub pid: usize,
+    pub cmdline: String,
+
+    inner_status: String,
+    inner_stat: String,
+
+    /// Processed `/proc/self/status` file
+    cached_status: Option<Rc<HashMap<String, String>>>,
+    /// Processed `/proc/self/stat` file
+    cached_stat: Option<Rc<Vec<String>>>,
```
cakebaker, line 169: Why do you store both the original and the processed values?

src/uu/pgrep/src/pgrep.rs
```diff
@@ -0,0 +1,321 @@
...
+        ));
+    }
+
+    // Verifying regex pattern
+    // And put it into static `REGEX`
+    if matches.get_flag("exact") {
+        let regex = Regex::new(pattern.first().unwrap())
+            .map_err(|e| USimpleError::new(2, e.to_string()))?;
+        REGEX.set(regex).unwrap();
+    }
```
cakebaker, line 64: I think there is a misunderstanding about regex patterns and `--exact`. Every pattern is a regex pattern. From the man page:
```
pgrep [options] pattern

[snipped]

pattern
              Specifies an Extended Regular Expression for matching against the process names or command lines.
```
And so something like `pgrep k(w|o)` is valid.

With `-x`/`--exact`, the pattern must match the entire process name. For example, `pgrep -x konsole` and `pgrep -x kons.*` return the PIDs of the `konsole` processes whereas `pgrep -x kons` doesn't return anything.

src/uu/pgrep/src/pgrep.rs
```diff
@@ -0,0 +1,321 @@
...
+#[uucore::main]
+pub fn uumain(args: impl uucore::Args) -> UResult<()> {
+    let matches = uu_app().try_get_matches_from(args)?;
+    let pattern = collect_arg_patterns(&matches);
+
+    // Pattern check
+    let flag_newest = matches.get_flag("newest");
+    let flag_oldest = matches.get_flag("oldest");
+
+    if (!flag_newest && !flag_oldest) && pattern.is_empty() {
```
cakebaker, line 44: This condition should also allow `--older` without a pattern because something like `pgrep --older=100` is valid.

src/uu/pgrep/src/pgrep.rs
```diff
@@ -0,0 +1,321 @@
...
+    }
+
+    // Collect pids
+    let pids = collect_matched_pids(&matches);
+
+    // Filtering pids
+    let pids = if flag_newest || flag_oldest {
+        let arg_older = matches.get_one::<u64>("older").unwrap_or(&0);
+        // `-o` and `-n` flag will be processed after pids collected
+        filter_oldest_newest(pids, flag_newest, *arg_older)
```
cakebaker, line 73: I already mentioned it above that something like `pgrep --older=100` is valid. That means it should work without `-o`/`-n`.

src/uu/pgrep/src/pgrep.rs
cakebaker, line 44: The same applies for `--runstates` because something like `pgrep --runstates=R` is valid.

src/uu/pgrep/src/pgrep.rs
```diff
@@ -0,0 +1,321 @@
...
+        if let Some(pid) = evaluate(pid) {
+            result.push(pid.clone())
+        }
+    }
+
+    result
+}
+
+// Make -o and -n as a group of args
+fn filter_oldest_newest(
```
cakebaker, line 211: I think this function is unnecessary. The list of PIDs you get from `collect_matched_pids` seems to be sorted (I'm not sure whether this is always the case) and so you can simply call `.first()`/`.last()` to get the oldest/newest PIDs.

src/uu/pgrep/src/process.rs
Krysztal112233: I was thinking of making a snapshot of all `/proc/<pid>` when the `pgrep` command is started, as there are times when slower processing may spawn new processes or some processes exit. Is this the right way to think about it?

src/uu/pgrep/src/pgrep.rs
Krysztal112233: So if I don't use the `-x` flag, will `pgrep` match cmdline or?

src/uu/pgrep/src/process.rs
cakebaker: Yes, that makes sense.

src/uu/pgrep/src/pgrep.rs
cakebaker: No, to match the cmdline you would have to use `-f`.

I think with `-x`, `pgrep` implicitly adds `^` and `$` to the regex. So the regexes from above become `^konsole$`, `^kons.*$`, and `^kons$`. The first two patterns match `konsole` and the third doesn't. If you omit the `^` and `$`, all three patterns match `konsole`.

tests/by-util/test_pgrep.rs
Krysztal112233: Hmmm ...... It looks like I didn't quite understand what you meant, could you please be more detailed?

cakebaker: Sure, what I wanted to say is that the output and exit code should be the same for `pgrep {(* --exact` and `pgrep {(*`. Currently, they are different.

Krysztal112233: I'll add it to the test later。

Krysztal112233: So I think it should be

```rust
#[test]
#[cfg(target_os = "linux")]
fn test_invalid_regex() {
    new_ucmd!().arg("{(*").arg("--exact").fails().code_is(2);
    new_ucmd!().arg("{(*").fails().code_is(2);
}
```

does it right?

cakebaker: That looks good :)

src/uu/pgrep/src/pgrep.rs
```diff
@@ -0,0 +1,288 @@
...
+        let mut pids = collect_matched_pids(&matches);
+        process_flag_o_n(&matches, &mut pids)
+    };
+
+    if pids.is_empty() {
+        uucore::error::set_exit_code(1);
+    }
+
+    // Processing output
+    let result = || {
```
tertsdiepraam, line 80: In Rust you can just do this:
```suggestion
    let output = {
```

tertsdiepraam: This seems to be mostly due to the newline, if you don't join with newlines, but put a newline after every entry (so at the end of each `format!`), it might become simpler.

tertsdiepraam (commented):
Small suggestion. I think @cakebaker is better equipped to do a review. Happy to help out more if you need it!

.gitignore
```diff
@@ -1 +1,2 @@
 target
+/.vscode/
```
cakebaker, line 2: I think this red symbol means there is a newline missing at the end of the file.

src/uu/pgrep/src/process.rs
```diff
@@ -0,0 +1,479 @@
...
+        })
+    }
+
+    pub fn inner_status(&self) -> &str {
+        &self.inner_status
+    }
+
+    pub fn inner_stat(&self) -> &str {
+        &self.inner_stat
+    }
```
cakebaker, line 226: I would rename those two functions because the "inner" in the function names is an implementation detail. The functions don't return an "inner status" or an "inner stat", they return the content of `/proc/<PID>/status` and `/proc/<PID>/stat`. And so my suggestion is to name them accordingly, something like `proc_status` and `proc_stat`. Or maybe you have better naming ideas ;-)

cakebaker (commented):
With the exception of the two function names it looks good.

src/uu/pgrep/src/process.rs
Krysztal112233: Cool, I think `proc_*` would be better

cakebaker:
Good work :)

PR #105, opened 2024-06-28
misc: Introduce CI workflow `Style/lint`

Krysztal112233:


.clippy.toml
```diff
@@ -0,0 +1 @@
+cognitive-complexity-threshold = 252
```
sylvestre, line 1: instead, please ignore in the code directly

sylvestre: #[allow(clippy::cognitive_complexity)]

Krysztal112233: Fixed

PR #112, opened 2024-07-08
pidof: Add command `pidof`

### LINKED ISSUE #93: Add command `pidof`
```
       pidof - find the process ID of a running program
```

Krysztal112233:
fix #93

### DIFF
--- a/src/uu/pidof/src/pidof.rs
+++ b/src/uu/pidof/src/pidof.rs
@@ -1,4 +1,5 @@
-use clap::{arg, crate_version, Arg, ArgAction, Command};
+use clap::{crate_version, Arg, ArgAction, ArgMatches, Command};
+use uu_pgrep::process::{walk_process, ProcessInformation};
 use uucore::{error::UResult, format_usage, help_about, help_usage};
 
 const ABOUT: &str = help_about!("pidof.md");
@@ -8,7 +9,61 @@ const USAGE: &str = help_usage!("pidof.md");
 pub fn uumain(args: impl uucore::Args) -> UResult<()> {
     let matches = uu_app().try_get_matches_from(args)?;
 
-    todo!()
+    let arg_program_name = matches.get_many::<String>("program-name");
+    let arg_separator = matches.get_one::<String>("d").unwrap();
+
+    if arg_program_name.is_none() {
+        uucore::error::set_exit_code(1);
+        return Ok(());
+    };
+
+    let mut collected = collect_matched_pids(&matches);
+    collected.sort_by(|a, b| b.pid.cmp(&a.pid));
+
+    let output = collected
+        .into_iter()
+        .map(|it| it.pid.to_string())
+        .collect::<Vec<_>>()
+        .join(arg_separator);
+
+    println!("{output}");
+
+    Ok(())
+}
+
+fn collect_matched_pids(matches: &ArgMatches) -> Vec<ProcessInformation> {
+    let program_name: Vec<_> = matches
+        .get_many::<String>("program-name")
+        .unwrap()
+        .cloned()
+        .collect();
+
+    let collected = walk_process().collect::<Vec<_>>();
+    let arg_omit_pid = matches
+        .get_many::<usize>("o")
+        .unwrap_or_default()
+        .copied()
+        .collect::<Vec<_>>();
+
+    let mut processed = Vec::new();
+    for mut process in collected {
+        let contains = program_name.contains(process.status().get("Name").unwrap());
+        let should_omit = arg_omit_pid.contains(&process.pid);
+
+        if contains && !should_omit {
+            processed.push(process)
+        }
+    }
+
+    let flag_s = matches.get_flag("s");
+    if flag_s {
+        match processed.first() {
+            Some(first) => vec![first.clone()],
+            None => Vec::new(),
+        }
+    } else {
+        processed
+    }
 }
 
 pub fn uu_app() -> Command {
@@ -25,12 +80,12 @@ pub fn uu_app() -> Command {
                 .required(true)
                 .index(1),
         )
-        .arg(
-            Arg::new("c")
-                .short('c')
-                .help("Return PIDs with the same root directory")
-                .action(ArgAction::SetTrue),
-        )
+        // .arg(
+        //     Arg::new("c")
+        //         .short('c')
+        //         .help("Return PIDs with the same root directory")
+        //         .action(ArgAction::SetTrue),
+        // )
         .arg(
             Arg::new("d")
                 .short('d')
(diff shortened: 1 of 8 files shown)


Krysztal112233:
Currently, `pgrep` is used as a temporary dependency because https://github.com/uutils/coreutils/pull/6483 is not merged and uutils is not releasing a new version.

cakebaker:
Which `pidof` version do you use as a reference? At least `pidof` from `procps-ng 4.0.4` doesn't contain some of the used options, for example `-z`.

Krysztal112233 (commented):
> Which `pidof` version do you use as a reference? At least `pidof` from `procps-ng 4.0.4` doesn't contain some of the used options, for example `-z`. 

From my apt output

```shell
❯ sudo apt install procps
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
procps is already the newest version (2:3.3.17-6ubuntu2.1).
```

Maybe it's too old?

sylvestre:
it is a bit old indeed
https://tracker.debian.org/pkg/procps
probably from an Ubuntu LTS ?

Krysztal112233:
> it is a bit old indeed https://tracker.debian.org/pkg/procps probably from an Ubuntu LTS ?

Yes I use Linuxmint21.3, based Ubuntu22.04.

src/uu/pidof/src/pidof.rs
```diff
@@ -0,0 +1,135 @@
+use clap::{crate_version, Arg, ArgAction, ArgMatches, Command};
```
sylvestre, line 1: please add the license header

tests/by-util/test_pidof.rs
```diff
@@ -0,0 +1,23 @@
...
+#[cfg(target_os = "linux")]
+fn test_find_init() {
+    new_ucmd!().arg("init").succeeds();
+}
+
+#[test]
+#[cfg(target_os = "linux")]
+fn test_find_kthreadd() {
+    new_ucmd!().arg("kthreadd").succeeds();
+}
```
cakebaker, line 23: These two tests test the happy path. What happens if you specify a name for which there is no PID?

tests/by-util/test_pidof.rs
```diff
@@ -0,0 +1,23 @@
...
+#[test]
+#[cfg(target_os = "linux")]
+fn test_find_init() {
+    new_ucmd!().arg("init").succeeds();
+}
+
+#[test]
+#[cfg(target_os = "linux")]
+fn test_find_kthreadd() {
+    new_ucmd!().arg("kthreadd").succeeds();
```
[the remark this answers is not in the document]
Krysztal112233: Humm I will reference to newer `pidof` later, it looks so weird.

Besides that, I noticed that the `pidof` command doesn't seem to be in the debian procps package now, why is that?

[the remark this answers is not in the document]
cakebaker: It's probably for historical reasons they use `pidof` from `sysvinit` (https://github.com/slicer69/sysvinit) and hence it is in the `sysvinit-utils` package. There is a long thread in their bug tracker about shipping `pidof` from `procps`: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=810018

src/uu/pidof/src/pidof.rs
```diff
@@ -0,0 +1,165 @@
...
+
+#[allow(clippy::cognitive_complexity)]
+pub fn uu_app() -> Command {
+    Command::new(uucore::util_name())
+        .version(crate_version!())
+        .about(ABOUT)
+        .override_usage(format_usage(USAGE))
+        .infer_long_args(true)
+        .disable_help_flag(true)
+        .disable_version_flag(true)
```
cakebaker, line 106: I know the `pidof` version you use as reference only supports `-h`. However, to be consistent with the other tools, I think it makes sense to support `--help`, `-V`, and `--version`, too.

src/uu/pidof/src/pidof.rs
```diff
@@ -0,0 +1,157 @@
...
+        //         .help("Avoid using stat system function on network shares")
+        //         .action(ArgAction::SetTrue),
+        // )
+        .arg(
+            Arg::new("o")
+                .short('o')
+                .help("Omit results with a given PID")
+                .action(ArgAction::Append)
+                .value_name("omitpid"),
+        )
```
cakebaker, line 138: I think you have to specify a value parser for this option. Currently I get an error when using `-o`:
```
cargo run pidof -o1 init
thread 'main' panicked at src/uu/pidof/src/pidof.rs:66:10:
Mismatch between definition and access of `o`. Could not downcast to usize, need to downcast to alloc::string::String
```

src/uu/pidof/src/pidof.rs
cakebaker, line 1: I noticed a bug related to `-s`. The original `pidof` returns the most recent PID:
```
# sleep 60 &
[1] 26
# sleep 60 &
[2] 27
# pidof sleep
27 26
# pidof -s sleep
27
```
Your implementation, on the other hand, returns the oldest PID:
```
$ sleep 60 &
[1] 16853
$ sleep 60 &
[2] 16854
$ cargo run pidof sleep
16854 16853
$ cargo run pidof -s sleep
16853
```

src/uu/pidof/src/pidof.rs
```diff
@@ -0,0 +1,161 @@
...
+    collected.sort_by(|a, b| b.pid.cmp(&a.pid));
+
+    let output = collected
+        .into_iter()
+        .map(|it| it.pid.to_string())
+        .collect::<Vec<_>>()
+        .join(arg_separator);
+
+    let flag_quite = matches.get_flag("q");
+    if !flag_quite {
```
cakebaker, line 43: A typo:
```suggestion
    let flag_quiet = matches.get_flag("q");
    if !flag_quiet {
```

src/uu/pidof/src/pidof.rs
cakebaker: I'm sorry, my advice was incorrect :| The original `pidof` also allows multiple PIDs separated by `,` and so a value parser doesn't work.
```
# pidof sleep
40 39 38
# pidof -o40,39 sleep
38
```

Krysztal112233: Humm I will take a look about it soon:P

src/uu/pidof/src/pidof.rs
```diff
@@ -0,0 +1,161 @@
...
+    Command::new(uucore::util_name())
+        .version(crate_version!())
+        .about(ABOUT)
+        .override_usage(format_usage(USAGE))
+        .infer_long_args(true)
+        .arg(
+            Arg::new("program-name")
+                .help("Program name.")
+                .required(true)
+                .index(1),
```
cakebaker, line 112: I think some setting is missing for this arg because currently I get an "unexpected argument" error when using multiple program names like `cargo run pidof bash bash bash`.

src/uu/pidof/src/pidof.rs
```diff
@@ -0,0 +1,161 @@
...
+    PathBuf::from(path)
+        .file_name()
+        .unwrap()
+        .to_str()
+        .unwrap()
+        .to_string()
+}
+
+fn collect_matched_pids(matches: &ArgMatches) -> Vec<ProcessInformation> {
+    let program_name: Vec<_> = matches
```
cakebaker, line 65: Here I would use the plural `program_names` because the variable is a `Vec` and thus can contain multiple program names.

cakebaker:
Thanks!

PR #123, opened 2024-07-14
pidof: Implemented `-o`

Krysztal112233:
fix #119 

This commit implemented two usage of `-o` which are already implemented in GNU's implementation, but not in ours.

```shell
❯ cargo run pidof zsh
116473 116472 116401 116391

❯ cargo run pidof -o116473,116472 zsh
116401 116391

❯ cargo run pidof -o116473 -o116472 zsh
116401 116391
```

And the GNU's implementation:

```shell
❯ pidof zsh
116473 116472 116401 116391

❯ pidof -o116473,116472 zsh
116401 116391

❯ pidof -o116473 -o116472 zsh
116401 116391
```

PR #124, opened 2024-07-14
pidof: Return multiple PIDs while multiple program

### LINKED ISSUE #118: pidof: When multiple program names are specified, `-s` should return multiple PIDs
The original `pidof` returns a PID for each specified program name whereas our `pidof` returns a single PID.
```
$ pidof -s bash bash bash
9024 9024 9024

$ cargo run pidof -s bash bash bash
9024
```

Krysztal112233:
fix #118

This commit fixed `-s` flag only output one result.

Before this commit:

```shell
❯ cargo run pidof -s zsh zsh zsh
218159
```

After:

```shell
❯ cargo run pidof -s zsh zsh zsh
218159 218159 218159
```

The behavior of uutils' implementation are as same as GNU's implementation now.

```shell
❯ cargo run pidof -s zsh zsh zsh
218159 218159 218159

❯ pidof -s zsh zsh zsh
218159 218159 218159
```

### DIFF
--- a/tests/by-util/test_pidof.rs
+++ b/tests/by-util/test_pidof.rs
@@ -46,13 +46,13 @@ fn test_s_flag() {
     let binding = new_ucmd!()
         .args(&["-s", "kthreadd", "kthreadd", "kthreadd"])
         .succeeds();
-    let output = binding.stdout_str();
+    let output = binding.stdout_str().trim_end();
 
-    let binding = output.replace('\n', "");
-    let pids = binding.split(' ').collect::<Vec<_>>();
+    let pids = output.split(' ').collect::<Vec<_>>();
     let first = pids[0];
 
     let result = pids.iter().all(|it| *it == first);
 
-    assert!(result)
+    assert!(result);
+    assert_eq!(pids.len(), 3);
 }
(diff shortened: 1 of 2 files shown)


tests/by-util/test_pidof.rs
```diff
@@ -33,3 +33,20 @@ fn test_no_program() {
...
     new_ucmd!().arg("NO_THIS_PROGRAM").fails().code_is(1);
 }
+
+#[test]
+#[cfg(target_os = "linux")]
+fn test_s_flag() {
+    let binding = new_ucmd!()
+        .args(&["-s", "kthreadd", "kthreadd", "kthreadd"])
+        .succeeds();
+    let output = binding.stdout_str();
```
cakebaker, line 43: I would just add a `trim` or `trim_end` here instead of the `replace` on line 45.

tests/by-util/test_pidof.rs
```diff
@@ -33,3 +33,20 @@ fn test_no_program() {
...
+
+    let binding = output.replace('\n', "");
+    let pids = binding.split(' ').collect::<Vec<_>>();
+    let first = pids[0];
+
+    let result = pids.iter().all(|it| *it == first);
+
+    assert!(result)
```
cakebaker, line 51: This assert passes without any of the changes made to `pidof.rs` ;-)  You also have to assert the number of PIDs.

cakebaker:
Looks good, though there is a conflict with the changes from #123 :|

cakebaker:
Thanks!

PR #131, opened 2024-07-18
pgrep: refactor pattern handling

cakebaker:
The motivation for this refactoring was to have everything related to pattern handling in a single function. For this purpose I renamed the function `collect_arg_patterns` to `try_get_pattern_from`, changed its return type from `Vec<String>` to `UResult<String>`, and moved some stuff from `uumain` to this function.


Krysztal112233 (approved):
Thanks!

PR #132, opened 2024-07-19
pgrep: allow `--older` without pattern

cakebaker:
This PR allows the `--older` option without a pattern: `pgrep --older=1`. This corresponds to the behavior of `pgrep` from procps-ng.


Krysztal112233 (approved):
Cool, thanks!

PR #138, opened 2024-07-20
pgrep: Add `TerminalType::Unknown` and implement `Display` for `TerminalType`

### LINKED ISSUE #136: pgrep: `--terminal` doesn't support `?` as value
`pgrep` from procps-ng supports `?` as a value for `--terminal` to match processes that do not have a controlling terminal. Our implementation doesn't support this value.
```
$ pgrep --terminal=? konsole
1265
1599
1640
1681
1741

$ cargo run pgrep --terminal=? konsole
// no output
```

Krysztal112233:
fix #136

### DIFF
--- a/src/uu/pgrep/src/process.rs
+++ b/src/uu/pgrep/src/process.rs
@@ -323,7 +323,11 @@ impl ProcessInformation {
 
         let path = PathBuf::from(format!("/proc/{}/fd", self.pid));
 
-        let mut result = fs::read_dir(path)?
+        let Ok(result) = fs::read_dir(path) else {
+            return Ok(Rc::new(HashSet::from_iter([TerminalType::Unknown])));
+        };
+
+        let mut result = result
             .flatten()
             .filter(|it| it.path().is_symlink())
             .flat_map(|it| fs::read_link(it.path()))
--- a/src/uu/pgrep/src/process.rs
+++ b/src/uu/pgrep/src/process.rs
@@ -17,13 +17,29 @@ pub enum TerminalType {
     Tty(u64),
     TtyS(u64),
     Pts(u64),
+    Unknown,
+}
+
+impl Display for TerminalType {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        match self {
+            Self::Tty(id) => write!(f, "/dev/pts/{}", id),
+            Self::TtyS(id) => write!(f, "/dev/tty{}", id),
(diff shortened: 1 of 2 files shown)


Krysztal112233:
Hmm... I'm not quite sure what kind of commands I should add to test, can you give some examples please?

cakebaker:
You could add something like `pgrep --terminal=? kthreadd` and `pgrep --terminal=tty1 kthreadd`. The first case should succeed and the other fail with your fix.

Krysztal112233:
> You could add something like `pgrep --terminal=? kthreadd` and `pgrep --terminal=tty1 kthreadd`. The first case should succeed and the other fail with your fix.

Nah I think this doesn't work on my system :(

```shell
$ pgrep --terminal=tty1 kthreadd
$ echo $?
1
$ pgrep --terminal=? kthreadd
$ echo $?
1
```

cakebaker:
Hm, are there any PIDs returned with `pgrep --terminal=?`?

Krysztal112233:
> Hm, are there any PIDs returned with `pgrep --terminal=?`?

After my investigation I realized that it was my procps version that was too low to support this syntax, and now I can get the output.

Now our implementation also works, but there are some problems: some high-privileged pid can't get to the terminal it belongs to, but the GNU implementation can.

```shell
❯ cargo run pgrep --terminal '?'
2977
2985
2986
2989
2992
2993
...
```

```shell
❯ pgrep --terminal=?
1
2
3
4
5
6
8
10
11
12
13
14
...
2977
2985
2986
2989
2992
2993
...
```

I don't know what to do about the above point for now, but for the `TerminalType` enumeration for `? ` support is now complete.

Krysztal112233:
And, is `TerminalType::Orphan` perhaps better than `TerminalType::Unknow`?

cakebaker:
I think `TerminalType::Unknown` is better.

src/uu/pgrep/src/process.rs
```diff
@@ -302,14 +323,18 @@ impl ProcessInformation {
...
 
-        let result = Rc::new(
-            fs::read_dir(path)?
-                .flatten()
-                .filter(|it| it.path().is_symlink())
-                .flat_map(|it| fs::read_link(it.path()))
-                .flat_map(TerminalType::try_from)
-                .collect::<HashSet<_>>(),
-        );
+        let mut result = fs::read_dir(path)?
```
cakebaker, line 326: I think you have to handle a permission error here and use a `TerminalType::Unknown` in such a case.

PR #139, opened 2024-07-22
pgrep: introduce `Settings` struct

cakebaker:
This PR introduces a `Settings` struct. The motivation for this refactoring was a separation of the user input handling from the other functionality.


Krysztal112233 (approved):
That's so cool, I think this pr will serve as a strong inspiration for my next implementations of other commands!

PR #143, opened 2024-07-24
pgrep: allow `--runstates` without pattern

cakebaker:
This PR allows the use of `--runstates` without a pattern. It also turns a regex pattern into a `const` in the tests.


Krysztal112233 (approved):
Thanks!

PR #144, opened 2024-07-25
pgrep: implement -P/--parent

cakebaker:


src/uu/pgrep/src/pgrep.rs
```diff
@@ -209,7 +214,20 @@ fn collect_matched_pids(settings: &Settings) -> Vec<ProcessInformation> {
             let arg_older = settings.older.unwrap_or(0);
             let older_matched = pid.start_time().unwrap() >= arg_older;
 
-            if (run_state_matched && pattern_matched && tty_matched && older_matched)
+            // the PPID is the fourth field in /proc/<PID>/stat
+            // (https://www.kernel.org/doc/html/latest/filesystems/proc.html#id10)
+            let stat = pid.stat();
+            let ppid = stat.get(3);
```
Krysztal112233, line 220: I noticed that the function `getppid` exists in libc, and in the future I'll add the corresponding wrapper to uucore, so could you please add a TODO here? Thank you!

Krysztal112233 (changes requested):

src/uu/pgrep/src/pgrep.rs
cakebaker: Hm, I don't think the `getppid` function from libc is useful in this context. It returns the PPID of the _current_ process according to the [docs](https://sourceware.org/glibc/manual/2.40/html_node/Process-Identification.html#index-getppid) whereas here the PPID of different processes is retrieved. Or am I misunderstanding something?

Krysztal112233: Oh, I think I misunderstood!

Krysztal112233:
Thanks :)

PR #166, opened 2024-08-07
pwdx: add support for multiple PIDs

cakebaker:
This PR adds support for multiple PIDs to match the behavior of the original `pwdx`.


Krysztal112233:
Thanks :)

PR #173, opened 2024-08-09
pwdx: use `sysinfo` crate

cakebaker:
This PR uses the `sysinfo` crate to get the working directories of processes in a platform-independent way.


Krysztal112233:
Thanks :)

PR #189, opened 2024-08-19
top: Add command `top`

### LINKED ISSUE #84: Add command `top`
could i implement ps and top for this, also there already exist projects like bottom

Krysztal112233:
Fix #84

### DIFF
--- a/src/uu/top/src/picker.rs
+++ b/src/uu/top/src/picker.rs
@@ -5,6 +5,10 @@
 
 use std::{
     ffi::OsString,
+    fs::File,
+    io::read_to_string,
+    path::PathBuf,
+    str::FromStr,
     sync::{OnceLock, RwLock},
 };
 use sysinfo::{Pid, System, Users};
@@ -157,10 +161,29 @@ fn command(pid: u32) -> String {
             .join(" ");
         let trimmed = binding.trim();
 
-        if trimmed.is_empty() {
-            "[kthreadd]".into()
+        let result: String = trimmed.into();
+
+        if cfg!(target_os = "linux") && result.is_empty() {
+            {
+                match PathBuf::from_str(&format!("/proc/{}/status", pid)) {
+                    Ok(path) => {
+                        let file = File::open(path).unwrap();
+                        let content = read_to_string(file).unwrap();
+                        let line = content
+                            .lines()
+                            .collect::<Vec<_>>()
+                            .first()
+                            .unwrap()
+                            .split(":")
+                            .collect::<Vec<_>>();
+
+                        line[1].trim().to_owned()
+                    }
+                    Err(_) => "".into(),
+                }
+            }
         } else {
-            trimmed.into()
+            result
         }
     };
 
--- a/src/uu/top/src/picker.rs
+++ b/src/uu/top/src/picker.rs
@@ -3,7 +3,7 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 
-use uu_pgrep::process::ProcessInformation;
+use uu_pgrep::process::{ProcessInformation, RunState};
 
 pub(crate) fn pickers(fields: &[String]) -> Vec<Box<dyn Fn(usize) -> String>> {
     fields
@@ -57,8 +57,10 @@ fn shr(_pid: usize) -> String {
     "TODO".into()
 }
 
-fn s(_pid: usize) -> String {
-    "TODO".into()
+fn s(pid: usize) -> String {
+    extractor(pid, |mut proc| {
+        proc.run_state().unwrap_or(RunState::Stopped).to_string()
+    })
 }
 
 fn time_plus(_pid: usize) -> String {
@@ -70,8 +72,16 @@ fn mem(_pid: usize) -> String {
 }
 
 fn command(pid: usize) -> String {
+    extractor(pid, |mut proc| proc.status()["Name"].clone())
+}
+
+/// If cannot constructing [ProcessInformation], it will return "?"
+fn extractor<F>(pid: usize, mut f: F) -> String
+where
+    F: FnMut(ProcessInformation) -> String,
+{
     match ProcessInformation::try_new(format!("/proc/{}/", pid).into()) {
-        Ok(mut proc) => proc.status()["Name"].clone(),
+        Ok(proc) => f(proc),
         Err(_) => "?".into(),
     }
 }
--- a/src/uu/top/src/picker.rs
+++ b/src/uu/top/src/picker.rs
@@ -3,6 +3,7 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 
+use chrono::{DateTime, NaiveDate, NaiveDateTime, NaiveTime};
 use std::{
     ffi::OsString,
     sync::{OnceLock, RwLock},
@@ -91,12 +92,34 @@ fn s(pid: u32) -> String {
         .to_string()
 }
 
-fn time_plus(_pid: u32) -> String {
-    "TODO".into()
+fn time_plus(pid: u32) -> String {
+    let binding = sysinfo().read().unwrap();
+    let Some(proc) = binding.process(Pid::from_u32(pid)) else {
+        return "0:00.00".into();
+    };
+
+    let (hour, min, sec) = {
+        let total = proc.run_time();
+        let hour = total / 3600;
+        let minute = (total % 3600) / 60;
+        let second = total % 60;
+
+        (hour, minute, second)
+    };
+
+    format!("{}:{:0>2}.{:0>2}", hour, min, sec)
 }
 
-fn mem(_pid: u32) -> String {
-    "TODO".into()
+fn mem(pid: u32) -> String {
+    let binding = sysinfo().read().unwrap();
+    let Some(proc) = binding.process(Pid::from_u32(pid)) else {
+        return "0.0".into();
+    };
+
+    format!(
+        "{:.1}",
+        proc.memory() as f32 / sysinfo().read().unwrap().total_memory() as f32
+    )
 }
 
 fn command(pid: u32) -> String {
--- a/src/uu/top/src/picker.rs
+++ b/src/uu/top/src/picker.rs
@@ -5,7 +5,6 @@
 
 use std::sync::{OnceLock, RwLock};
 use sysinfo::{Pid, System};
-use uu_pgrep::process::{ProcessInformation, RunState};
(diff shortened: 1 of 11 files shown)


Krysztal112233:
This pr only implemented output, and the TUI will implemented in other pr as same as `slabtop`

Krysztal112233:
There are more problems with the current pr, but solutions do not yet exist for these problems.

- There should be a word like `rt` in `PR` field, but the getpriority function does not provide a corresponding response.
- The value of the `PR` field looks incorrect, the `top` command output should be 20 mostly, but the current implementation mostly outputs 0 or 19.

cakebaker:
I guess you have seen that `test_flag_user` fails on Windows?

tests/by-util/test_top.rs
```diff
@@ -0,0 +1,28 @@
...
+
+#[test]
+fn test_conflict_arg() {
+    new_ucmd!().arg("-p 0").arg("-U 0").fails().code_is(1);
+}
+
+#[test]
+fn test_flag_user() {
+    let binding = new_ucmd!().arg("-U=root").succeeds();
+    let output = binding.code_is(0).stderr_str();
```
cakebaker, line 21: Why do you use `stderr_str()` and not `stdout_str()`?

src/uu/top/src/top.rs
```diff
@@ -0,0 +1,239 @@
...
+
+        Self { filter, width }
+    }
+}
+
+#[uucore::main]
+pub fn uumain(args: impl uucore::Args) -> UResult<()> {
+    let matches = uu_app().try_get_matches_from(args)?;
+
+    // Must refresh twice.
```
cakebaker, line 60: Why?

src/uu/top/src/top.rs
```diff
@@ -0,0 +1,239 @@
...
+    fn new(matches: &ArgMatches) -> Self {
+        let filter = if let Some(pidlist) = matches.get_many::<u32>("pid") {
+            Some(Filter::Pid(pidlist.cloned().collect::<Vec<_>>()))
+        } else if let Some(user) = matches.get_one::<String>("filter-any-user") {
+            Some(Filter::User(user.clone()))
+        } else {
+            matches
+                .get_one::<String>("filter-only-euser")
+                .map(|euser| Filter::EUser(euser.clone()))
+        };
```
cakebaker, line 48: It looks a bit odd to use two different ways to accomplish the same. Maybe it would make sense to add another `if let`.
```rust
        } else if let Some(user) = matches.get_one::<String>("filter-any-user") {
            Some(Filter::User(user.clone()))
        } else if let Some(euser) = matches.get_one::<String>("filter-only-euser") {
            Some(Filter::EUser(euser.clone()))
        } else {
            None
        };
```
Another idea is to get rid of all the `if`s and use `map` and `or_else`.
```rust
        let filter = matches
            .get_many::<u32>("pid")
            .map(|pidlist| Filter::Pid(pidlist.cloned().collect()))
            .or_else(|| {
                matches
                    .get_one::<String>("filter-any-user")
                    .map(|user| Filter::User(user.clone()))
            })
            .or_else(|| {
                matches
                    .get_one::<String>("filter-only-euser")
                    .map(|euser| Filter::EUser(euser.clone()))
            });
```

src/uu/top/src/top.rs
```diff
@@ -0,0 +1,239 @@
...
+    let settings = Settings::new(&matches);
+
+    if let Some(Filter::User(user)) = &settings.filter {
+        if Users::new_with_refreshed_list()
+            .iter()
+            .all(|it| it.name() != user)
+        {
+            return Err(USimpleError::new(1, "top: Invalid user"));
+        }
+    }
```
cakebaker, line 74: This works only partially. When using `--filter-any-user=0`, an "invalid user" error is shown whereas the original `top` shows no error because it treats numbers as uids.

The handling of `--filter-only-euser` should probably be similar to the handling of `--filter-any-user`.

src/uu/top/src/top.rs
```diff
@@ -0,0 +1,239 @@
...
+    picker::sysinfo().write().unwrap().refresh_all();
+
+    let settings = Settings::new(&matches);
+
+    if let Some(Filter::User(user)) = &settings.filter {
+        if Users::new_with_refreshed_list()
+            .iter()
+            .all(|it| it.name() != user)
+        {
+            return Err(USimpleError::new(1, "top: Invalid user"));
```
cakebaker, line 72: `USimpleError` automatically prefixes the error message with the tool name, so currently the error message shown is "top: top: Invalid user".
```suggestion
            return Err(USimpleError::new(1, "Invalid user"));
```

src/uu/top/src/top.rs
Krysztal112233: If we want to fetch CPU usage of process, we have to fresh twice, it's required by sysinfo crate.

[sysinfo](https://docs.rs/sysinfo/0.31.2/sysinfo/struct.System.html#method.refresh_cpu_usage)

tests/by-util/test_top.rs
```diff
@@ -0,0 +1,41 @@
...
+        new_ucmd!()
+            .arg("-U=root")
+            .succeeds()
+            .code_is(0)
+            .stdout_str(),
+    );
+    check(new_ucmd!().arg("-U=0").succeeds().code_is(0).stdout_str());
+
+    new_ucmd!().arg("-U=NOT_EXIST").fails().code_is(1);
+    new_ucmd!().arg("-U=19999").fails().code_is(1);
```
cakebaker, line 40: I think this is not correct. When running `-U=19999` with the original `top`, it doesn't fail. It simply shows the header and no processes.

src/uu/top/src/top.rs
cakebaker: Can you please add this link to the comment?

src/uu/top/src/top.rs
```diff
@@ -0,0 +1,275 @@
...
+            // arg!(-d  --delay                <SECS>          "iterative delay as SECS [.TENTHS]"),
+            // arg!(-E  --"scale-summary-mem"  <SCALE>         "set mem as: k,m,g,t,p,e for SCALE"),
+            // arg!(-e  --"scale-task-mem"     <SCALE>         "set mem with: k,m,g,t,p for SCALE"),
+            // arg!(-H  --"threads-show"                       "show tasks plus all their threads"),
+            // arg!(-i  --"idle-toggle"                        "reverse last remembered 'i' state"),
+            // arg!(-n  --iterations           <NUMBER>        "exit on maximum iterations NUMBER"),
+            arg!(-O  --"list-fields"                        "output all field names, then exit"),
+            // arg!(-o  --"sort-override"      <FIELD>         "force sorting on this named FIELD"),
+            arg!(-p  --pid                  <PIDLIST>       "monitor only the tasks in PIDLIST")
+                .action(ArgAction::Append),
```
cakebaker, line 266: I think a value parser is missing here. When running `cargo run -q top -p <PID>` I get the following error:
```
thread 'main' panicked at src/uu/top/src/top.rs:41:14:
Mismatch between definition and access of `pid`. Could not downcast to u32, need to downcast to alloc::string::String
```

src/uu/top/src/picker.rs
```diff
@@ -0,0 +1,177 @@
...
+    let f = |cmd: &[OsString]| -> String {
+        let binding = cmd
+            .iter()
+            .map(|os_str| os_str.to_string_lossy().into_owned())
+            .collect::<Vec<_>>()
+            .join(" ");
+        let trimmed = binding.trim();
+
+        if trimmed.is_empty() {
+            "[kthreadd]".into()
```
cakebaker, line 161: Why do you use `[kthreadd]`? The original `top` doesn't use `[kthreadd]`:
```
26     root              0    TODO  TODO  TODO  TODO  S  0.00   0.0   8:12.49  [kthreadd] 
57     root              0    TODO  TODO  TODO  TODO  S  0.00   0.0   8:12.49  [kthreadd]

26 root     -51   0       0      0      0 S   0.0   0.0   0:00.00 idle_inject/2                                                                               
57 root     -51   0       0      0      0 S   0.0   0.0   0:01.55 irq/9-acpi
```

tests/by-util/test_top.rs
Krysztal112233: Fixed 

```shell
❯ cargo -q run top -U1999
TODO


 PID  USER  PR  NI  VIRT  RES  SHR  S  %CPU  %MEM  TIME+  COMMAND 
```

src/uu/top/src/top.rs
Krysztal112233: Fixed 

```shell
❯ cargo run -q top -p 1
TODO


 PID  USER  PR  NI    VIRT  RES   SHR   S  %CPU  %MEM  TIME+    COMMAND 
 1    root  0   TODO  TODO  TODO  TODO  S  0.00  0.0   2:19.31  /sbin/init splash 
```

src/uu/top/src/top.rs
```diff
@@ -0,0 +1,272 @@
...
+
+#[derive(Debug)]
+struct Settings {
+    // batch:bool
+    filter: Option<Filter>,
+    width: Option<usize>,
+}
+
+impl Settings {
+    fn new(matches: &ArgMatches) -> UResult<Self> {
```
cakebaker, line 39: Why did you change the return type from `Self` to `UResult<Self>`?

src/uu/top/src/top.rs
```diff
@@ -0,0 +1,272 @@
...
+            .or_else(|| {
+                matches
+                    .get_one::<String>("filter-only-euser")
+                    .map(|euser| Filter::EUser(euser.clone()))
+            });
+
+        let filter = match filter {
+            Some(Filter::User(data)) => Some(Filter::User(try_into_uid(data)?)),
+            // TODO: Make sure this working
+            Some(Filter::EUser(data)) => Some(Filter::User(try_into_uid(data)?)),
```
cakebaker, line 79: I guess it should be `EUser` in both cases:
```suggestion
            Some(Filter::EUser(data)) => Some(Filter::EUser(try_into_uid(data)?)),
```

src/uu/top/src/top.rs
cakebaker, line 1: I'm not sure whether you are aware of it: the recent changes broke `--filter-any-user`. When running `cargo run top --filter-any-user=root`, it no longer shows any processes.

Krysztal112233: Yes, I'll fix it

PR #193, opened 2024-08-19
pmap: fix address column

cakebaker:
This PR fixes the output of the address column when running `cargo run pmap <some PID>` and shows the start address only.

Before the change:
```
...
7addfaf2d000-7addfaf2f000 rw-p 00036000 08:08 10773308                   /usr/lib/ld-linux-x86-64.so.2
...
```
After the change:
```
...
00007addfaf2d000 rw-p 00036000 08:08 10773308                   /usr/lib/ld-linux-x86-64.so.2
...
```
In comparison, the output of the original `pmap`:
```
...
00007addfaf2d000      8K rw--- ld-linux-x86-64.so.2
...
```
The PR also removes an unnecessary header and adds two additional spaces to the output of the first line.


Krysztal112233:
Thanks :)

PR #197, opened 2024-08-21
pmap: fix perms column & remove unnecessary columns

cakebaker:
This PR fixes the "perms" column and removes unnecessary columns from the output when running `cargo run pmap <some PID>`.

Before the changes:
```
0000735e54227000      8K rw-p 00036000 08:08 10773308                   /usr/lib/ld-linux-x86-64.so.2
```
After the changes:
```
0000735e54227000      8K rw--- /usr/lib/ld-linux-x86-64.so.2
```
In comparison, the output of the original `pmap`:
```
0000735e54227000      8K rw--- ld-linux-x86-64.so.2
```


Krysztal112233 (approved):
Thanks :)

PR #201, opened 2024-08-25
pmap: fix "filename" column

cakebaker:
This PR fixes the output of the "filename" column when running `cargo run pmap <PID>`. The output of the memory map should now match the output of the original `pmap`.

Before the changes:
```
000075ce972a3000      8K rw--- /usr/lib/ld-linux-x86-64.so.2
00007ffe29309000    132K rw--- [stack]
ffffffffff600000      4K --x-- [vsyscall]
```

After the changes:
```
000075ce972a3000      8K rw--- ld-linux-x86-64.so.2
00007ffe29309000    132K rw---   [ stack ]
ffffffffff600000      4K --x--   [ anon ]
```


Krysztal112233 (approved):
Great work :)

PR #202, opened 2024-08-26
pmap: trim cmdline

cakebaker:
This PR trims the return value of the `parse_cmdline` function.

Before the change, there is a space at the end of the output:
```
$ cargo run -q pmap 1 | tr " " "."
1:.../sbin/init.\vmlinuz-linux.
```
After the change:
```
$ cargo run -q pmap 1 | tr " " "."
1:.../sbin/init.\vmlinuz-linux
```


Krysztal112233:
Hmmm ...... See `pmap` up not working correctly?

```shell
❯ pmap 1 | tr " " "."
1:.../sbin/init.splash

❯ cargo run -q pmap 1 | tr " " "."
pmap:.function/utility.not.found
```

cakebaker:
Hm, stupid question: are you running it in the correct project? What is shown if you omit the `-q` param?

PR #206, opened 2024-08-28
pidof: adapt options to those of `pidof` from `procps-ng`

cakebaker:
I thought a bit about `pidof` and I think our implementation should match the `pidof` from `procps-ng`, and not the one from `sysvinit` that comes with Debian (there is a pending [proposal](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=810018) to switch to the `pidof` from `procps-ng`).

And so this PR adapts the options to those of the `pidof` from `procps-ng`. The functionality of the options seems to be the same, so no changes there. 

Before the changes:
```
Options:
  -d <sep>          Use the provided character as output separator
  -o <omitpid>      Omit results with a given PID
  -q                Quiet mode. Do not display output
  -s                Only return one PID
  -h, --help        Print help
  -V, --version     Print version
```

After the changes:
```
Options:
  -S, --separator <SEP>  Use SEP as separator between PIDs
  -o, --omit-pid <PID>   Omit results with a given PID
  -q                     Quiet mode. Do not display output
  -s, --single-shot      Only return one PID
  -h, --help             Print help
  -V, --version          Print version
```

### DIFF
--- a/src/uu/pidof/src/pidof.rs
+++ b/src/uu/pidof/src/pidof.rs
@@ -156,6 +156,7 @@ pub fn uu_app() -> Command {
         .arg(
             Arg::new("s")
                 .short('s')
+                .long("single-shot")
                 .help("Only return one PID")
                 .action(ArgAction::SetTrue),
         )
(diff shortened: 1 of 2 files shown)


Krysztal112233:
But I think maybe we can keep the `-d` flag until for the proposal approved?

cakebaker:
Changes since last push:
* added `-d` as an alias after the suggestion from @Krysztal112233

Krysztal112233:
Thanks, great work:)

Krysztal112233 (approved):

PR #239, opened 2024-10-06
readme: add `hugetop` & `sysctl` to todo list

cakebaker:
I noticed that procps-ng contains two utilities we don't mention in our readme: `hugetop` (a new utility that will come with the next release) and `sysctl`. And so this PR adds them to our todo list.


sylvestre:
hugetop, it is a cute name :)
could you please open them as good first issue ? :)

PR #249, opened 2024-10-09
pmap: implement `--device`

cakebaker:
This PR implements `-d`/`--device` and creates output in the following format:
```
13176:   ../hello_world/target/debug/hello_world
Address           Kbytes Mode  Offset           Device    Mapping
000060ab2984a000      24 r---- 0000000000000000 008:00008 hello_world
000060ab29850000     248 r-x-- 0000000000006000 008:00008 hello_world
...
ffffffffff600000       4 --x-- 0000000000000000 000:00000   [ anon ]
mapped: 3060K    writeable/private: 348K    shared: 0K
```

### DIFF
--- a/src/uu/pmap/src/maps_format_parser.rs
+++ b/src/uu/pmap/src/maps_format_parser.rs
@@ -5,12 +5,15 @@
 
 use std::io::{Error, ErrorKind};
 
-// Represents a parsed single line from /proc/<PID>/maps.
+// Represents a parsed single line from /proc/<PID>/maps for the default and device formats. It
+// omits the inode information because it's not used by those formats.
 #[derive(Debug, PartialEq)]
 pub struct MapLine {
     pub address: String,
     pub size_in_kb: u64,
     pub perms: String,
+    pub offset: String,
+    pub device: String,
     pub mapping: String,
 }
 
@@ -32,7 +35,18 @@ pub fn parse_map_line(line: &str) -> Result<MapLine, Error> {
         .ok_or_else(|| Error::from(ErrorKind::InvalidData))?;
     let perms = parse_perms(perms);
 
-    let mapping: String = rest.splitn(4, ' ').skip(3).collect();
+    let (offset, rest) = rest
+        .split_once(' ')
+        .ok_or_else(|| Error::from(ErrorKind::InvalidData))?;
+    let offset = format!("{offset:0>16}");
+
+    let (device, rest) = rest
+        .split_once(' ')
+        .ok_or_else(|| Error::from(ErrorKind::InvalidData))?;
+    let device = parse_device(device)?;
+
+    // skip the "inode" column
+    let mapping: String = rest.splitn(2, ' ').skip(1).collect();
     let mapping = mapping.trim_ascii_start();
     let mapping = parse_mapping(mapping);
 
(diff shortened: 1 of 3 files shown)


src/uu/pmap/src/pmap.rs
```diff
@@ -88,7 +89,47 @@ fn parse_maps(pid: &str) -> Result<u64, Error> {
         total += map_line.size_in_kb;
     }
 
-    Ok(total)
+    println!(" total {total:>16}K");
+
+    Ok(())
+}
+
+fn output_device_format(pid: &str) -> Result<(), Error> {
```
sylvestre, line 97: Seems that there are some duplication here.

What about doing something like:
```
fn process_maps<F>(pid: &str, mut process_line: F) -> Result<(), Error>
where
    F: FnMut(&MapLine),
{
    let path = format!("/proc/{pid}/maps");
    let contents = fs::read_to_string(path)?;
    
    for line in contents.lines() {
        let map_line = parse_map_line(line);
        process_line(&map_line);
    }
    
    Ok(())
}


```

and 
```
fn output_default_format(pid: &str) -> Result<(), Error> {
    let mut total = 0;
    process_maps(pid, |map_line| {
        total += map_line.size_in_kb;
    })?;
    println!(" total {total:>16}K");
    Ok(())
}
```

src/uu/pmap/src/maps_format_parser.rs
```diff
@@ -47,13 +99,13 @@ fn parse_address(memory_range: &str) -> (String, u64) {
...
-    let perms = perms.replace("p", "-");
-
-    // the fifth char seems to be always '-' in the original pmap
-    format!("{perms}-")
+// Pads the device info from /proc/<PID>/maps with zeros and turns AB:CD into 0AB:000CD.
+fn parse_device(device: &str) -> String {
+    if let Some((major, minor)) = device.split_once(':') {
+        format!("{major:0>3}:{minor:0>5}")
+    } else {
+        unreachable!("device must contain a ':'")
```
sylvestre, line 107: unreachable will panic

maybe just something like
        Err(Error::new(ErrorKind::InvalidData, "Invalid device format"))

src/uu/pmap/src/pmap.rs
```diff
@@ -49,11 +49,12 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
             }
         }
 
-        match parse_maps(pid) {
-            Ok(total) => println!(" total {total:>16}K"),
-            Err(_) => {
-                set_exit_code(1);
-            }
+        if matches.get_flag(options::DEVICE) {
+            output_device_format(pid).map_err(|_| set_exit_code(1)).ok();
```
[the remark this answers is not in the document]
cakebaker: Yes, the original `pmap` doesn't show an error message if `/proc/<PID>/maps` can't be read:
```
$ pmap 1 --device
1:   /sbin/init \vmlinuz-linux
$ echo $?
1
```
In this example, no memory map is shown because `/proc/1/maps` is not accessible for a normal user.

src/uu/pmap/src/maps_format_parser.rs
cakebaker: Hm. I wrote `maps_format_parser.rs` with the assumption that the value passed to `parse_map_line()` has the correct format ;-)  It's the reason it returns a `MapLine` and not a `Result` and there are also some `expect`s that will panic. So should I write it more defensively?

src/uu/pmap/src/pmap.rs
cakebaker: Good point, thanks!

PR #250, opened 2024-10-10
watch: Windows support

xbjfk:
This commit adds Windows support to `watch`. To run commands, cmd.exe /C is used. In tests, rundll32.exe is used instead of true.


cakebaker:
Thanks for your PR!

PR #251, opened 2024-10-10
free: add Windows support

xbjfk:
Values are virtual the same as resmon. Unfortunately, it is slightly hacky to get the page "Swap" file sizes.

I'm not sure about the name/location of the `windows_util.rs` file, let me know if you have a better name for that.

NOTE: I've noticed that the main coreutils crate uses `windows-sys` instead of `windows`. If this is something that needs to be done here, let me know.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -877,7 +877,7 @@ dependencies = [
  "memchr",
  "ntapi",
  "rayon",
- "windows",
+ "windows 0.57.0",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


cakebaker:
> NOTE: I've noticed that the main coreutils crate uses `windows-sys` instead of `windows`. If this is something that needs to be done here, let me know.

I think it is fine to use `windows` instead of `windows-sys`. In `coreutils`, there was an attempt to switch to `windows`, though the PR never left the draft status:|

cakebaker:
Thanks!

PR #292, opened 2024-12-30
feat: tui for top command

r3yc0n1c:
### Changes Made:

- [x] Dynamically pulled all other data from the main top.rs file
- [x] Added an exit feature by pressing 'q'
- [x] Implemented a static top block (pending implementation in top.rs)

### To Do:
- [ ] Implement data refresh functionality for the table and the top block

https://github.com/user-attachments/assets/fc9e05a9-40ce-464f-b147-303ebe2a0904

PR #306, opened 2025-01-13
top: implement SUMMARY display

### LINKED ISSUE #207: Implement SUMMARY display for `top`
Currently our implementation output:

```
TODO

```

But the GNU's implementation:

```
top - 15:51:38 up 9 min,  1 user,  load average: 1.45, 1.09, 0.61
Tasks: 496 total,   1 running, 495 sleeping,   0 stopped,   0 zombie
%Cpu(s):  3.5 us,  1.7 sy,  0.0 ni, 94.8 id,  0.0 wa,  0.0 hi,  0.0 si,  0.0 st 
MiB Mem :  31506.4 total,  20616.9 free,   5700.0 used,   5692.4 buff/cache     
MiB Swap:      0.0 total,      0.0 free,      0.0 used.  25806.5 avail Mem 

```

Bluemangoo:
Resolves #207 

Tasks:
- [x] time and uptime
- [x] active user count
- [ ] cpu load average
- [x] task summary
- [ ] cpu load
- [x] memory and swap summary
- [x] determine memory unit from `--scale-summary-mem`

I add crate `systemstat` for some of the values.

```text
top - 13:03:00 up 1:33, 1 user, load average: 2.01, 1.48, 1.28
Tasks: 1040 total, 6 running, 974 sleeping, 0 stopped, 0 zombie
%Cpu(s):  13.9 us, 3.3 sy, 0.0 ni, 82.4 id, 0.3 wa, 0.0 hi, 0.1 si, 0.0 st
MiB Mem :   7636.4 total,    433.3 free,   6415.1 used,    788.0 buff/cache
MiB Swap:  30518.0 total,  27218.5 free,   3299.5 used,   1221.3 avail Mem
```

### DIFF
--- a/src/uu/top/src/header.rs
+++ b/src/uu/top/src/header.rs
@@ -35,11 +35,6 @@ extern "C" {
     ) -> libc::c_int;
 }
 
-#[cfg(target_os = "macos")]
-fn todo() -> String {
-    "TODO".into()
-}
-
 fn format_memory(memory_b: u64, unit: u64) -> f64 {
     ByteSize::b(memory_b).0 as f64 / unit as f64
 }
@@ -55,6 +50,7 @@ pub fn get_nusers_systemd() -> uucore::error::UResult<usize> {
     use uucore::error::USimpleError;
     use uucore::libc::*;
 
+    // SAFETY: sd_booted to check if system is booted with systemd.
     unsafe {
         // systemd
         if sd_booted() > 0 {
@@ -210,25 +206,28 @@ fn cpu() -> String {
         interrupt_time: 0,
         interrupt_count: 0,
     };
+    // SAFETY: malloc is safe to use here. We free the memory after we are done with it. If action fails, all "time" will be 0.
     unsafe {
         let len = n_cpu * size_of::<SystemProcessorPerformanceInformation>();
         let data = malloc(len);
-        let _ = NtQuerySystemInformation(
+        let status = NtQuerySystemInformation(
             windows_sys::Wdk::System::SystemInformation::SystemProcessorPerformanceInformation,
             data,
             (n_cpu * size_of::<SystemProcessorPerformanceInformation>()) as u32,
             std::ptr::null_mut(),
         );
-        for i in 0..n_cpu {
-            let cpu = data.add(i * size_of::<SystemProcessorPerformanceInformation>())
-                as *const SystemProcessorPerformanceInformation;
-            let cpu = cpu.as_ref().unwrap();
-            cpu_load.idle_time += cpu.idle_time;
-            cpu_load.kernel_time += cpu.kernel_time;
-            cpu_load.user_time += cpu.user_time;
-            cpu_load.dpc_time += cpu.dpc_time;
-            cpu_load.interrupt_time += cpu.interrupt_time;
-            cpu_load.interrupt_count += cpu.interrupt_count;
+        if status == 0 {
+            for i in 0..n_cpu {
+                let cpu = data.add(i * size_of::<SystemProcessorPerformanceInformation>())
+                    as *const SystemProcessorPerformanceInformation;
+                let cpu = cpu.as_ref().unwrap();
+                cpu_load.idle_time += cpu.idle_time;
+                cpu_load.kernel_time += cpu.kernel_time;
+                cpu_load.user_time += cpu.user_time;
+                cpu_load.dpc_time += cpu.dpc_time;
+                cpu_load.interrupt_time += cpu.interrupt_time;
+                cpu_load.interrupt_count += cpu.interrupt_count;
+            }
         }
     }
     let total = cpu_load.idle_time
@@ -249,7 +248,7 @@ fn cpu() -> String {
 //TODO: Implement for macos
 #[cfg(target_os = "macos")]
 fn cpu() -> String {
-    todo()
+    "TODO".into()
 }
 
 fn memory(scale_summary_mem: Option<&String>) -> String {
--- a/src/uu/top/src/header.rs
+++ b/src/uu/top/src/header.rs
@@ -7,38 +7,6 @@ use crate::picker::{sysinfo, systemstat};
 use bytesize::ByteSize;
 use clap::ArgMatches;
 use systemstat::Platform;
-#[cfg(not(any(target_os = "macos", target_os = "linux")))]
-use {
-    std::sync::{Mutex, OnceLock},
-    systemstat::{CPULoad, DelayedMeasurement},
-};
-
-#[cfg(not(any(target_os = "macos", target_os = "linux")))]
-static LOAD_AVERAGE: OnceLock<Mutex<DelayedMeasurement<CPULoad>>> = OnceLock::new();
-
-#[cfg(not(any(target_os = "macos", target_os = "linux")))]
-pub(crate) fn cpu_load() -> CPULoad {
-    match LOAD_AVERAGE.get() {
-        None => {
-            LOAD_AVERAGE.get_or_init(|| {
-                Mutex::new(systemstat().read().unwrap().cpu_load_aggregate().unwrap())
-            });
-            systemstat()
-                .read()
-                .unwrap()
-                .cpu_load_aggregate()
-                .unwrap()
-                .done()
-                .unwrap()
-        }
-        Some(v) => {
-            let mut v = v.lock().unwrap();
-            let load = v.done().unwrap();
-            *v = systemstat().read().unwrap().cpu_load_aggregate().unwrap();
-            load
-        }
-    }
-}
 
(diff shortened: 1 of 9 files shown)


Krysztal112233:
That's great. But can I ask if it can you use text instead of image? That would be more search friendly.

Bluemangoo:
There are no way to measure load average on Windows. We can only get an *instantaneous* CPU load with this command: 

```text
wmic cpu get loadpercentage
```

The Python wheel `psutil` emulates what Linux does (for sure, need time to collect data):

https://github.com/giampaolo/psutil/blob/d4b37d9628e634c00a2990e488b1cc38ea6a41b5/psutil/arch/windows/wmi.c#L16-L54

I'm not sure what should we do: ignore it, or do what `psutil` does.

Krysztal112233:
In fact, the crate we used required us does the same thing:

```rust
    // Must refresh twice.
    // https://docs.rs/sysinfo/0.31.2/sysinfo/struct.System.html#method.refresh_cpu_usage
    picker::sysinfo().write().unwrap().refresh_all();
    sleep(Duration::from_millis(200));
    picker::sysinfo().write().unwrap().refresh_all();
```

You can even use a multi-threaded background to refresh this data if `sysinfo` cannot doesn't meet your needs.

src/uu/top/src/header.rs
```diff
@@ -0,0 +1,347 @@
+use crate::picker::{sysinfo, systemstat};
```
sylvestre, line 1: please add the license header to this file (header.rs ;)

Bluemangoo: resolved.

Bluemangoo:
i add `libsystemd-sys` to get correct active users count(on linux).

```
starting with systemd 256.7-2, utmp support has been disabled. This
was done as utmp is not Y2038 safe.

After this change, /run/utmp is no longer created by systemd
```

i check if system is booted by systemd, if so get active user count from it, or from utmp, like what gnu does.

so these apt package `libsystemd-dev` is required. shall i install it in workflows?

Krysztal112233:
> i add `libsystemd-sys` to get correct active users count(on linux).
> 
> ```
> starting with systemd 256.7-2, utmp support has been disabled. This
> was done as utmp is not Y2038 safe.
> 
> After this change, /run/utmp is no longer created by systemd
> ```
> 
> i check if system is booted by systemd, if so get active user count from it, or from utmp, like what gnu does.
> 
> so the apt package `libsystemd-dev` is required. shall i install it in workflows?

Is there any other way to detect the system is initialized with systemd? (e.g. some features on the `Init` process?)

Bluemangoo:
Yes i used #[cfg].

https://github.com/Bluemangoo/procps/blob/9b45c481c1450371f2c9f3c54ab4e64ad290e335/src/uu/top/src/header.rs#L157

so as the deps.

https://github.com/Bluemangoo/procps/blob/9b45c481c1450371f2c9f3c54ab4e64ad290e335/src/uu/top/Cargo.toml#L25

if system is booted by systemd, `sd_booted` will return a number bigger than 0.

https://github.com/Bluemangoo/procps/blob/9b45c481c1450371f2c9f3c54ab4e64ad290e335/src/uu/top/src/header.rs#L165

Bluemangoo:
I would like to mention that [platform-specific dependencies](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#platform-specific-dependencies) do not work with workspace dependencies. Therefore, I specify the version of these dependencies in top's `Cargo.toml` file, rather than that in the root.

Bluemangoo:
The way CPU time is calculated in Windows differs from that in Linux. On Windows, CPU time is divided into the following categories: `idle`, `kernel`, `user`, `dpc`, and `interrupt`. The `dpc` time is **similar** to Linux's `soft interrupt`, while the `interrupt` time is specifically for hardware. 

Should I display them in the following format:  
`%Cpu(s):  0.7 us,  50.0 sy,  49.3 id,  0.0 hi,  0.0 si`  
or in this format:  
`%Cpu(s):  0.7 us,  50.0 sy,  49.3 id,  0.0 hi,  0.0 dpc`?

Additionally, in the [Windows documentation](https://learn.microsoft.com/en-us/windows/win32/api/winternl/nf-winternl-ntquerysysteminformation#system_processor_performance_information), it mentions reserved fields (`dpc`, `interrupt`) in the struct:

```c
typedef struct
_SYSTEM_PROCESSOR_PERFORMANCE_INFORMATION {
    LARGE_INTEGER IdleTime;
    LARGE_INTEGER KernelTime;
    LARGE_INTEGER UserTime;
    LARGE_INTEGER Reserved1[2];
    ULONG Reserved2;
} SYSTEM_PROCESSOR_PERFORMANCE_INFORMATION;
```

I found a possible explanation in [psutil](https://github.com/giampaolo/psutil/blob/a509e5aa1829c0268cd4d069ac340a5d9fb4fee8/psutil/arch/windows/ntextapi.h#L149-L161).

sylvestre:
please let me know when I can review it. It doesn't have to be completed to land!

Bluemangoo:
It seems that the load average on Windows may require opening a new issue. The CPU load for macOS is still to be done.

you can begin reviewing the rest of the content.

src/uu/top/src/header.rs
```diff
@@ -0,0 +1,362 @@
...
+#[cfg(not(target_os = "linux"))]
+fn todo() -> String {
+    "TODO".into()
+}
+
+fn format_memory(memory_b: u64, unit: u64) -> f64 {
+    ByteSize::b(memory_b).0 as f64 / unit as f64
+}
+
+fn uptime() -> String {
```
sylvestre, line 36: seems very similar to this function:
https://github.com/uutils/coreutils/blob/main/src/uu/uptime/src/uptime.rs#L420

what about creating a function in uucore and using it in both places?

src/uu/top/src/header.rs
```diff
@@ -0,0 +1,362 @@
...
+    Utmpx::iter_all_records().for_each(|ut| {
+        if ut.record_type() == 7 && !ut.user().is_empty() {
+            num_user += 1;
+        }
+    });
+    format_user(num_user)
+}
+
+#[cfg(not(target_os = "windows"))]
+fn load_average() -> String {
```
sylvestre, line 175: maybe we could use some elements from
https://github.com/uutils/coreutils/blob/main/src/uu/uptime/src/uptime.rs#L239
?

src/uu/top/src/header.rs
```diff
@@ -0,0 +1,362 @@
...
+    )
+}
+
+//TODO: Implement for macos
+#[cfg(target_os = "macos")]
+fn cpu() -> String {
+    todo()
+}
+
+fn memory(arg: &ArgMatches) -> String {
```
sylvestre, line 334: instead of passing ArgMatches, you could just pass arg.get_one::<String>("scale-summary-mem")

would be easier to understand + easier to test this function :)

src/uu/top/src/header.rs
sylvestre: note that we can do that after it lands :)

Bluemangoo: uu_uptime just print the first line of top summary.

Shall I move all these three functions to uu_core, and in uu_uptime just call these functions instead of re-implementation?

Bluemangoo: I have implemented some of the TODOs in `uu_uptime`, and now I need the OpenBSD implementation in `uu_uptime`. I believe these can be integrated together.

sylvestre: > Shall I move all these three functions to uu_core, and in uu_uptime just call these functions instead of re-implementation?

correct (if the current tests passes)

Bluemangoo:
all fixed.

src/uu/top/src/header.rs
```diff
@@ -0,0 +1,283 @@
...
+                as *const SystemProcessorPerformanceInformation;
+            let cpu = cpu.as_ref().unwrap();
+            cpu_load.idle_time += cpu.idle_time;
+            cpu_load.kernel_time += cpu.kernel_time;
+            cpu_load.user_time += cpu.user_time;
+            cpu_load.dpc_time += cpu.dpc_time;
+            cpu_load.interrupt_time += cpu.interrupt_time;
+            cpu_load.interrupt_count += cpu.interrupt_count;
+        }
+    }
```
Krysztal112233, line 233: Could you add `// SAFETY` to explain why you use `unsafe` here and why it's safety here?

https://github.com/uutils/coreutils/blob/main/CONTRIBUTING.md#unsafe

Bluemangoo:
not severe issue, you can also merge first, and update later. I hope #292 is not waiting for this PR.

PR #308, opened 2025-01-17
sysctl: Add tool

dezgeg:
This supports basic get and set of sysctls on Linux. Main missing thing is reading /etc/sysctl.d files.

### DIFF
--- a/src/uu/sysctl/Cargo.toml
+++ b/src/uu/sysctl/Cargo.toml
@@ -1,0 +1,25 @@
+[package]
+name = "uu_sysctl"
+version = "0.0.1"
+edition = "2021"
+authors = ["uutils developers"]
+license = "MIT"
+description = "sysctl ~ (uutils) Show or modify kernel parameters at runtime"
+
+homepage = "https://github.com/uutils/procps"
+repository = "https://github.com/uutils/procps/tree/main/src/uu/sysctl"
+keywords = ["acl", "uutils", "cross-platform", "cli", "utility"]
+categories = ["command-line-utilities"]
+
+[dependencies]
+uucore = { workspace = true }
+clap = { workspace = true }
+sysinfo = { workspace = true }
+walkdir = { workspace = true }
+
+[lib]
+path = "src/sysctl.rs"
+
+[[bin]]
+name = "sysctl"
+path = "src/main.rs"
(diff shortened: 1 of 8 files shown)


src/uu/sysctl/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+name = "uu_sysctl"
+version = "0.0.1"
+edition = "2021"
+authors = ["uutils developers"]
+license = "MIT"
+description = "sysctl ~ (uutils) Report current working directory of a process"
```
sylvestre, line 7: I think it is incorrect :)

src/uu/sysctl/src/sysctl.rs
```diff
@@ -0,0 +1,191 @@
...
+            match get_sysctl(&var) {
+                Ok(v) => v,
+                Err(e) => {
+                    if !matches.get_flag("ignore") {
+                        uucore::show_error!("error reading key '{}': {}", var, e);
+                        set_exit_code(1);
+                    }
+                    continue;
+                }
+            }
```
sylvestre, line 107:
```suggestion
        get_sysctl(&var).unwrap_or_else(|e| {
    if !matches.get_flag("ignore") {
        uucore::show_error!("error reading key '{}': {}", var, e);
        set_exit_code(1);
    }
    return;
});
```

tests/tests.rs
```diff
@@ -56,3 +56,7 @@ mod test_snice;
 #[cfg(feature = "pkill")]
 #[path = "by-util/test_pkill.rs"]
 mod test_pkill;
+
+#[cfg(feature = "sysctl")]
+#[path = "by-util/test_sysctl.rs"]
```
sylvestre, line 61: would be nice to add a few tests :)

src/uu/sysctl/Cargo.toml
dezgeg: Oops, fixed

tests/tests.rs
dezgeg: duh... `git add`ed now.

src/uu/sysctl/src/sysctl.rs
dezgeg: I don't think I can do `continue` from inside the closure (to skip the print).. I guess splitting a function (that processes single var[=value]) ?

dezgeg: Should be better now

sylvestre:
LGTM, thanks :)

PR #318, opened 2025-02-06
pidof: Support '-t' flag

dezgeg:
This flag makes pidof print thread ids of threads belonging to the matching processes.

### DIFF
--- a/src/uu/pgrep/src/process.rs
+++ b/src/uu/pgrep/src/process.rs
@@ -186,6 +186,8 @@ pub struct ProcessInformation {
     cached_stat: Option<Rc<Vec<String>>>,
 
     cached_start_time: Option<u64>,
+
+    cached_thread_ids: Option<Rc<Vec<usize>>>,
 }
 
 impl ProcessInformation {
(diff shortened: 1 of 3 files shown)


sylvestre:
Could you please add a test to make sure we don't regress? Thanks

dezgeg:
> Could you please add a test to make sure we don't regress? Thanks

Added

cakebaker (commented):
Looks good.

I noticed a small difference to the original `pidof`, though: it lists the IDs in a descending order whereas we use an ascending order:
```
$ pidof konsole -t
26231 26230 26229 26228 1365 1364 1363 1362 1361
$ cargo run -q  pidof konsole -t
1361 1362 1363 1364 1365 26228 26229 26230 26231
```
Without the `-t` flag, both use a descending order.

dezgeg:
> Looks good.
> 
> I noticed a small difference to the original `pidof`, though: it lists the IDs in a descending order whereas we use an ascending order:
> 
> ```
> $ pidof konsole -t
> 26231 26230 26229 26228 1365 1364 1363 1362 1361
> $ cargo run -q  pidof konsole -t
> 1361 1362 1363 1364 1365 26228 26229 26230 26231
> ```
> 
> Without the `-t` flag, both use a descending order.

Good catch, also found out that `-s`  combined with `-t` wasn't working right. Both should work now.

Krysztal112233 (approved):
Great work :)

PR #321, opened 2025-02-11
Unify matching logic in pgrep and pkill

dezgeg:
Currently pgrep and pkill implement slightly different set command line args for process matching, even though both commands should support the same features. This PR extends the functionality of both tools to their superset of features. This will allow deduplicating the process matching part into a common module in a future PR.

### DIFF
--- a/src/uu/pgrep/src/pgrep.rs
+++ b/src/uu/pgrep/src/pgrep.rs
@@ -10,6 +10,8 @@ use clap::{arg, crate_version, Arg, ArgAction, ArgGroup, ArgMatches, Command};
 use process::{walk_process, ProcessInformation, Teletype};
 use regex::Regex;
 use std::{collections::HashSet, sync::OnceLock};
+#[cfg(unix)]
+use uucore::{display::Quotable, signals::signal_by_name_or_value};
 use uucore::{
     error::{UResult, USimpleError},
     format_usage, help_about, help_usage,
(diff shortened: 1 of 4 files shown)


tests/by-util/test_pgrep.rs
```diff
@@ -347,3 +347,23 @@ fn test_parent_non_matching_parent() {
...
 }
+
+#[test]
+#[cfg(target_os = "linux")]
+fn test_require_handler() {
+    new_ucmd!()
+        .arg("--require-handler")
+        .arg("--signal=INT")
+        .arg("NONEXISTENT")
+        .fails();
```
cakebaker, line 358: A detail: I would also ensure that there is no output.
```suggestion
        .fails()
        .no_output();
```

src/uu/pgrep/src/pgrep.rs
```diff
@@ -275,6 +289,17 @@ fn process_flag_o_n(
...
+#[cfg(unix)]
+fn parse_signal_value(signal_name: &str) -> UResult<usize> {
+    let optional_signal_value = signal_by_name_or_value(signal_name);
+    match optional_signal_value {
+        Some(x) => Ok(x),
+        None => Err(USimpleError::new(
+            1,
+            format!("Unknown signal {}", signal_name.quote()),
+        )),
+    }
```
cakebaker, line 301: You can simplify this block by using `ok_or_else`:
```suggestion
    optional_signal_value
        .ok_or_else(|| USimpleError::new(1, format!("Unknown signal {}", signal_name.quote())))
```

tests/by-util/test_pgrep.rs
dezgeg: Good point, added.

src/uu/pgrep/src/pgrep.rs
dezgeg: Indeed, that is better. Changed.

cakebaker:
Thanks!

PR #322, opened 2025-02-13
pgrep, pkill: Unify process matching code 

dezgeg:
After previous changes the process matching part in pgrep and pkill are identical. Extract it into a common module for both.


cakebaker:
Nice refactoring, thanks!

PR #328, opened 2025-02-18
process_matcher: Fix accidentally matching against pid

### LINKED ISSUE #307: pkill: Some behavior seems doesn't match
```
➜  procps git:(pkill) sleep 60&
[2] 529099
➜  procps git:(pkill) cargo run pkill 529099
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.08s
     Running `target/debug/procps pkill 529099`
[2]  + 529099 terminated  sleep 60
```

```
➜  procps git:(pkill) sleep 60&
[1] 530352
➜  procps git:(pkill) pkill 530352
➜  procps git:(pkill)
```

_Originally posted by @Krysztal112233 in https://github.com/uutils/procps/issues/290#issuecomment-2579427832_

dezgeg:
Currently pgrep/pkill without -x/-f flags is matching based on the first 15 characters of /proc/<pid>/stat, which actually contains something like "1116878 (cat) R" thus matching the process id when it should just match on the name.

This has probably come from misunderstanding the comment from manpage:

> The process name used for matching is limited to the 15 characters
> present in the output of /proc/pid/stat.

... which doesn't mean pgrep/pkill is literally matching on 15 characters of /proc/<pid>/stat but that the process name in that file is truncated to 15 characters.

Fixes #307


sylvestre:
would it be possible to add a test to make sure we don't regress? thanks

dezgeg:
> would it be possible to add a test to make sure we don't regress? thanks

Added

cakebaker:
Thanks!

PR #332, opened 2025-02-23
pgrep/pkill/pidwait: Support --uid, --euid & --group options

dezgeg:


src/uu/pgrep/src/process.rs
```diff
@@ -296,6 +296,43 @@ impl ProcessInformation {
...
+        self.get_uid_or_gid_field("Uid", 1)
+    }
+
+    pub fn gid(&mut self) -> Result<u32, io::Error> {
+        self.get_uid_or_gid_field("Gid", 0)
+    }
+
+    pub fn egid(&mut self) -> Result<u32, io::Error> {
+        self.get_uid_or_gid_field("Gid", 1)
+    }
```
Krysztal112233, line 334: Hello, there's a plan for migrating `process.rs` into `uucore` and those function already implemented in `uucore` :)

https://github.com/uutils/coreutils/blob/604cc404ff1ffa4a1b8dc66e3bc29b7b14b0619c/src/uucore/src/lib/features/process.rs#L21-L44

Krysztal112233: The current implementation of `process.rs` is only for Linux, and will need to be implemented for other *NIX systems in the future (maybe Windows too?).

Krysztal112233 (commented):
Could you sync the changes of `process.rs` into https://github.com/uutils/coreutils/blob/main/src/uucore/src/lib/features/process.rs? This file will be migrated into `uucore` in future :)

dezgeg:
> Could you sync the changes of `process.rs` into https://github.com/uutils/coreutils/blob/main/src/uucore/src/lib/features/process.rs? This file will be migrated into `uucore` in future :)

Added here: https://github.com/uutils/coreutils/pull/7351

Is there some timeline for the migration? Probably makes sense to get the process module in procps as complete as possible before migration, as it will make considerably more difficult to add new features to procps (as then new features might need to wait for uucore to make a release)

Krysztal112233:
> Is there some timeline for the migration?

If possible, the migration will start when tools that depend on `pgrep` are almost available. Actually I don't know the timeline but it was planned last year by me :(. So just continue your work and I will make pull request to do this job before migration.

cakebaker:
@dezgeg thanks!

Regarding moving `process.rs` to `coreutils`: currently I don't see what the benefit is. And I agree with @dezgeg that it would make working on this project rather inflexible.

PR #339, opened 2025-02-26
process_matcher: Don't match own procps process

dezgeg:
For example, pgrep should never report itself as a match (but it can match other pgrep processes).


cakebaker:
Good catch, thanks :)

PR #343, opened 2025-02-28
pgrep: Support -w option

dezgeg:
Unlike pidof's similar option, the pgrep version will run the match predicates against every thread of the system and printing matching tids (unlike pidof which will run match predicates on every process of the system and then list every thread of every matched process).


cakebaker:
Good work, thanks!

PR #360, opened 2025-03-21
pgrep/pkill/pidwait: --require-handler fixes

dezgeg:
See commit messages

### DIFF
--- a/tests/by-util/test_pgrep.rs
+++ b/tests/by-util/test_pgrep.rs
@@ -382,6 +382,15 @@ fn test_signal_that_never_matches() {
 
 #[test]
 #[cfg(target_os = "linux")]
+fn test_lone_require_handler_is_allowed() {
+    new_ucmd!()
+        .arg("--require-handler")
+        .run()
+        .stderr_does_not_contain("no matching criteria specified");
+}
+
+#[test]
+#[cfg(target_os = "linux")]
 fn test_does_not_match_pid() {
     let our_pid = std::process::id();
     new_ucmd!().arg(our_pid.to_string()).fails();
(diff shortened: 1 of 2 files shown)


tests/by-util/test_pgrep.rs
```diff
@@ -369,6 +369,34 @@ fn test_invalid_signal() {
...
+        .fails()
+        .no_output();
+}
+
+#[test]
+#[cfg(target_os = "linux")]
+fn test_only_require_handler() {
+    new_ucmd!()
+        .arg("--require-handler")
+        .arg("--signal=0")
```
cakebaker, line 395: What's the reason for this argument? Isn't this test about using `--require-handler` alone?

tests/by-util/test_pgrep.rs
```diff
@@ -369,6 +369,34 @@ fn test_invalid_signal() {
...
 
+#[test]
+#[cfg(target_os = "linux")]
+fn test_signals_that_never_match() {
+    new_ucmd!()
+        .arg("--require-handler")
+        .arg("--signal=0")
+        .arg(".*")
+        .fails()
+        .no_output();
```
cakebaker, line 380: I think the assumption that the `0` signal never matches is incorrect. If I run the command with the original `pgrep`, and I have `top` running, it returns the PID of `top`.

tests/by-util/test_pgrep.rs
dezgeg: It is there to make the exit status consistently 0, regardless of what else is running on the system. I guess alternative would be to ignore the exit status but just check that stderr doesn't contain the usage error

cakebaker: I assumed it's the role of `--inverse` to ensure the exit status is consistently 0.

dezgeg: Changed

Krysztal112233 (approved):

Issue #183, opened 2024-08-16
w: shows more users than `w` from procps-ng


cakebaker:
Our implementation of `w` shows more users than `w` from procps-ng:
```
$ cargo run -q w
USER     TTY      LOGIN@   IDLE   JCPU   PCPU WHAT
dho      tty1     07:29    TODO   0.08   0.08 /usr/bin/startplasma-wayland
dho      pts/0    07:29    TODO   5127.583.13 /usr/bin/kded6
dho      pts/1    07:29    TODO   87.50  0.01 /bin/bash
dho      pts/2    07:32    TODO   1.95   0.05 /bin/bash
dho      pts/3    07:32    TODO   0.36   0.24 /bin/bash
dho      pts/4    07:32    TODO   0.19   0.04 /bin/bash
dho      pts/5    07:32    TODO   0.06   0.06 /bin/bash

$ w
 16:31:58 up  9:03,  2 users,  load average: 0.26, 0.24, 0.20
USER     TTY       LOGIN@   IDLE   JCPU   PCPU  WHAT
dho      tty1      07:29    9:03m  0.08s  0.08s /usr/bin/startplasma-wayland
dho                07:29    8:59m  0.00s  0.95s /usr/lib/systemd/systemd --user
 ```

**Comment by aisuneko** [comment]
I think this is because our implementation fetches the list of users via utmpx (utmp/wtmp login records):
```
     Running `target/debug/procps w`
USER     TTY      LOGIN@   IDLE   JCPU   PCPU WHAT
aisuneko tty2     Sun23    10:18m 0.22   0.2  /usr/bin/startplasma-wayland
aisuneko pts/0    Sun23    10:18m 1303.853.35 /usr/bin/kded6
aisuneko pts/2    Sun23    10:13m 0.46   0.37 /usr/bin/fish
aisuneko pts/3    Sun23    10:13m 0.46   0.09 sudotio/dev/ttyACM0
aisuneko pts/4    Sun23    10:10m 0.04   0.04 /usr/bin/fish
aisuneko pts/1    09:48    37.05s 0.15   0.15 /usr/bin/fish
[aisuneko@liquidice procps]$ last -f /var/run/utmp
aisuneko pts/1        :0               Mon Mar 24 09:48   still logged in
aisuneko pts/4        :0               Sun Mar 23 23:40   still logged in
aisuneko pts/3        :0               Sun Mar 23 23:40   still logged in
[...]

PR #361, opened 2025-03-22
w: implement idle_time and elapsed time formatting

aisuneko:
This is a basic implementation of the display of each users' idle time, as well as a dedicated time formatting function `format_time_elapsed` for this feature, as my first PR to the `uutils` project.

~~The code's logic is largely based on the original source of procps~~ We'll, no, as that is considered license violation. I mean just the logic is similar, but the code is rather another implementation.

### DIFF
--- a/src/uu/w/src/w.rs
+++ b/src/uu/w/src/w.rs
@@ -3,20 +3,16 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 
+use std::time::Duration;
+
 #[cfg(target_os = "linux")]
 use chrono::Datelike;
 use clap::crate_version;
 use clap::{Arg, ArgAction, Command};
 #[cfg(target_os = "linux")]
 use libc::{sysconf, _SC_CLK_TCK};
-use std::process;
 #[cfg(target_os = "linux")]
-use std::{
-    collections::HashMap,
-    fs,
-    path::Path,
-    time::{Duration, SystemTime},
-};
+use std::{collections::HashMap, fs, path::Path, process, time::SystemTime};
 #[cfg(target_os = "linux")]
 use uucore::utmpx::Utmpx;
 use uucore::{error::UResult, format_usage, help_about, help_usage};
@@ -99,7 +95,11 @@ fn fetch_idle_time(tty: String) -> Result<Duration, std::io::Error> {
     }
 }
 
-#[cfg(target_os = "linux")]
+#[cfg(not(target_os = "linux"))]
+fn fetch_idle_time(tty: String) -> Result<Duration, std::io::Error> {
+    Ok(Duration::ZERO)
+}
+
 fn format_time_elapsed(
     time: Duration,
     old_style: bool,
@@ -121,16 +121,14 @@ fn format_time_elapsed(
             t.num_seconds() % 60,
             if old_style { "m" } else { "" }
         ))
+    } else if old_style {
+        Ok(String::new())
     } else {
-        if old_style {
-            Ok(format!(""))
-        } else {
-            Ok(format!(
-                "{}.{:02}s",
-                t.num_seconds() % 60,
-                (t.num_milliseconds() % 1000) / 10
-            ))
-        }
+        Ok(format!(
+            "{}.{:02}s",
+            t.num_seconds() % 60,
+            (t.num_milliseconds() % 1000) / 10
+        ))
     }
 }
 
--- a/src/uu/w/src/w.rs
+++ b/src/uu/w/src/w.rs
@@ -189,7 +189,7 @@ fn fetch_user_info() -> Result<Vec<UserInfo>, std::io::Error> {
 
 #[cfg(any(target_os = "macos", target_os = "windows"))]
 fn fetch_user_info() -> Result<Vec<UserInfo>, std::io::Error> {
-    Ok(Vec::new())process, 
+    Ok(Vec::new())
 }
 
 #[uucore::main]
--- a/src/uu/w/src/w.rs
+++ b/src/uu/w/src/w.rs
@@ -3,8 +3,6 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 
-use std::time::Duration;
-
 #[cfg(target_os = "linux")]
 use chrono::Datelike;
 use clap::crate_version;
@@ -12,7 +10,8 @@ use clap::{Arg, ArgAction, Command};
 #[cfg(target_os = "linux")]
 use libc::{sysconf, _SC_CLK_TCK};
 #[cfg(target_os = "linux")]
-use std::{collections::HashMap, fs, path::Path, process, time::SystemTime};
+use std::{collections::HashMap, fs, path::Path, time::SystemTime};
+use std::{process, time::Duration};
 #[cfg(target_os = "linux")]
 use uucore::utmpx::Utmpx;
 use uucore::{error::UResult, format_usage, help_about, help_usage};
@@ -96,7 +95,7 @@ fn fetch_idle_time(tty: String) -> Result<Duration, std::io::Error> {
 }
 
 #[cfg(not(target_os = "linux"))]
-fn fetch_idle_time(tty: String) -> Result<Duration, std::io::Error> {
+fn fetch_idle_time(_tty: String) -> Result<Duration, std::io::Error> {
     Ok(Duration::ZERO)
 }
 
@@ -190,7 +189,7 @@ fn fetch_user_info() -> Result<Vec<UserInfo>, std::io::Error> {
 
 #[cfg(any(target_os = "macos", target_os = "windows"))]
 fn fetch_user_info() -> Result<Vec<UserInfo>, std::io::Error> {
-    Ok(Vec::new())
+    Ok(Vec::new())process, 
 }
 
 #[uucore::main]
--- a/src/uu/w/src/w.rs
+++ b/src/uu/w/src/w.rs
@@ -11,7 +11,12 @@ use clap::{Arg, ArgAction, Command};
 use libc::{sysconf, _SC_CLK_TCK};
 use std::process;
 #[cfg(target_os = "linux")]
-use std::{collections::HashMap, fs, path::Path};
+use std::{
+    collections::HashMap,
+    fs,
+    path::Path,
+    time::{Duration, SystemTime},
+};
 #[cfg(target_os = "linux")]
 use uucore::utmpx::Utmpx;
 use uucore::{error::UResult, format_usage, help_about, help_usage};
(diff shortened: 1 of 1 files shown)


Krysztal112233:
Nah... Out project licensed under MIT and the member who want to contributing **_CAN'T SEE_** the GPL licensed code:

> Generally, we try to follow what GNU is doing in terms of options and behavior. It is recommended to look at the GNU coreutils manual (on the web, or locally using info <utility>). It is more in depth than the man pages and provides a good description of available features and their implementation details. But remember, you cannot look at the GNU source code!

from our https://github.com/uutils/coreutils/blob/main/CONTRIBUTING.md#writing-code

Krysztal112233:
And very thanks for your contributing, but we need to discuss those code will merge into our code base or not

Could you please give us some help?

@sylvestre @cakebaker

sylvestre:
please don't look at the other implementation. Licenses aren't compatible.
However, here, as you didn't copy how the code is done or its structure, I think we are good

aisuneko:
> please don't look at the other implementation. Licenses aren't compatible.

Sorry about that. I didn't notice that the licenses are incompatible... I'll pay extra attention to that in future contributions.
(Though IMHO: if we are to "follow what GNU is doing in terms of options and behavior", and in particular, *to pass GNU tests*, then I think looking at the source code is inevitable to some point (that is to better understand how it works/study it's output because we need to follow it's behavior, not to directly copy it from C to Rust)

sylvestre:
Looking at test and code are different things.

In general, if you need to look at code, look at the BSD implementation.
for example:
https://github.com/apple-oss-distributions/shell_cmds/blob/main/w/w.c

aisuneko:
> Looking at test and code are different things.

Then what is our stance on the tests and the scripts/code that might appear in these tests? E.g. Are we allowed to port and run the tests in our project?

> In general, if you need to look at code, look at the BSD implementation. for example: https://github.com/apple-oss-distributions/shell_cmds/blob/main/w/w.c

Thanks for the explanation.

Krysztal112233:
> Then what is our stance on the tests and the scripts/code that might appear in these tests? E.g. Are we allowed to port and run the tests directly in our project?

The test suite will only check that the inputs, outputs and behaviors are as expected, it only provides the usage of util, not the original logic of the util implemented and licensed by GNU and GPL, so it is possible to see the test suite's code

Krysztal112233:
And don't forget to fix the CI failures, that's a great job :)

aisuneko:
> > Then what is our stance on the tests and the scripts/code that might appear in these tests? E.g. Are we allowed to port and run the tests directly in our project?
> 
> The test suite will only check that the inputs, outputs and behaviors are as expected, it only provides the usage of util, not provide the original logic of the util implemented and licensed by GNU and GPL, so it is possible to check the test suite's code

Thanks for clarifying. I might work toward porting the tests to the rest of the uutils projects (as I'm also interested in contributing to uutils for this year's GSoC), so do let me know if possible license infringements exist.

> And don't forget to fix the CI failures, that's a great job :)

I suppose that's due to my current implementation is for Linux only which causes the Windows and macOS runners to fail. I don't always have Windows or macOS machines available, so suggestions on ways of porting my code to these two platforms are appreciated ;)

Krysztal112233:
Hum you can see the logs of failed CI, the logs will provided all the things you need to fix CI failure

src/uu/w/src/w.rs
```diff
@@ -83,6 +88,52 @@ fn fetch_pcpu_time(pid: i32) -> Result<f64, std::io::Error> {
...
+    let path = Path::new("/dev/").join(tty);
+    let stat = fs::metadata(path)?;
+    if let Ok(time) = stat.accessed() {
+        Ok(SystemTime::now().duration_since(time).unwrap_or_default())
+    } else {
+        Ok(Duration::ZERO)
+    }
+}
+
+#[cfg(target_os = "linux")]
```
aisuneko, line 102: ... where this directive right here should be the only cause for the problem

Krysztal112233 (commented):
from CI `Style/lint (windows-latest)` and `Style/lint (macos-latest)`

```rust
 error[E0433]: failed to resolve: use of undeclared crate or module `process`
   --> src\uu\w\src\w.rs:241:13
    |
241 |             process::exit(1);
    |             ^^^^^^^ use of undeclared crate or module `process`
    |
help: consider importing this module
    |
6   + use std::process;
    |

error: unused variable: `tty`
  --> src\uu\w\src\w.rs:99:20
   |
99 | fn fetch_idle_time(tty: String) -> Result<Duration, std::io::Error> {
   |                    ^^^ help: if this is intentional, prefix it with an underscore: `_tty`
   |
   = note: `-D unused-variables` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(unused_variables)]`
```

please fix this error, this error caused build fail :)

Krysztal112233:
Cool, the last step you have to do:

```rust
 error: function `fetch_idle_time` is never used
  --> src/uu/w/src/w.rs:98:4
   |
98 | fn fetch_idle_time(_tty: String) -> Result<Duration, std::io::Error> {
   |    ^^^^^^^^^^^^^^^
   |
   = note: `-D dead-code` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(dead_code)]`
```

Krysztal112233:
does there any output of `w` command on your environment? I cannot get any output on my environment

```shell
➜  procps git:(main) cargo run -q w  
USER     TTY      LOGIN@   IDLE   JCPU   PCPU WHAT
➜  procps git:(main) w
 15:57:56 up  4:05,  2 users,  load average: 1.45, 2.01, 1.88
USER     TTY      FROM             LOGIN@   IDLE   JCPU   PCPU WHAT
krysztal tty1     -                11:54    4:05m  0.09s  0.09s /usr/bin/startplasma-wayland
krysztal          -                11:54     ?     0.00s  9.93s /usr/lib/systemd/systemd --user
```

aisuneko:
@Krysztal112233 It works as intended on my machine, as in https://github.com/uutils/procps/issues/183#issuecomment-2746681917 :
```
[aisuneko@liquidice procps]$ cargo run -q w
USER     TTY      LOGIN@   IDLE   JCPU   PCPU WHAT
aisuneko tty2     Sun23    41:43m 0.22   0.2  /usr/bin/startplasma-wayland
aisuneko pts/0    Sun23    41:43m 11244.5023.31/usr/bin/kded6
aisuneko pts/2    Sun23    41:39m 0.46   0.37 /usr/bin/fish
aisuneko pts/3    Sun23    41:39m 0.46   0.09 sudotio/dev/ttyACM0
aisuneko pts/4    Sun23    17:06m 0.07   0.07 /usr/bin/fish
aisuneko pts/1    Mon24    4:40m  0.25   0.25 /usr/bin/fish
[aisuneko@liquidice procps]$ w
 17:20:25 up 1 day, 17:44,  1 user,  load average: 0.48, 0.58, 0.66
USER     TTY       LOGIN@   IDLE   JCPU   PCPU  WHAT
aisuneko tty2      Sun23   41:44m  0.22s  0.20s /usr/bin/startplasma-wayland
[aisuneko@liquidice procps]$ 


```

Krysztal112233:
Weird problem. Our implementation output empty content on Debian.

```shell
➜  procps git:(aisuneko/main) cargo run -q w -V
w 0.0.1
➜  procps git:(aisuneko/main) w -V
w from procps-ng 4.0.4
```

But please don't worry, it's not your problem. :)

I'll spend some time researching why, this pr looks perfect.

Can you add some tests to this file? https://github.com/uutils/procps/blob/main/tests/by-util/test_w.rs

cakebaker:
@Krysztal112233 I get a similar output as @aisuneko on my machine using Arch Linux.

src/uu/w/src/w.rs
```diff
@@ -83,6 +83,54 @@ fn fetch_pcpu_time(pid: i32) -> Result<f64, std::io::Error> {
...
+        Ok(SystemTime::now().duration_since(time).unwrap_or_default())
+    } else {
+        Ok(Duration::ZERO)
+    }
+}
+
+#[cfg(not(target_os = "linux"))]
+fn _fetch_idle_time(_tty: String) -> Result<Duration, std::io::Error> {
+    Ok(Duration::ZERO)
+}
```
cakebaker, line 100: The function is never called, so it is not necessary:
```suggestion
```

src/uu/w/src/w.rs
```diff
@@ -23,7 +23,7 @@ struct UserInfo {
     user: String,
     terminal: String,
     login_time: String,
-    idle_time: String,
+    idle_time: Duration, // for better compatiability with old-style outputs
```
cakebaker, line 26: A typo:
```suggestion
    idle_time: Duration, // for better compatibility with old-style outputs
```

src/uu/w/src/w.rs
```diff
@@ -83,6 +83,54 @@ fn fetch_pcpu_time(pid: i32) -> Result<f64, std::io::Error> {
...
+
+#[cfg(not(target_os = "linux"))]
+fn _fetch_idle_time(_tty: String) -> Result<Duration, std::io::Error> {
+    Ok(Duration::ZERO)
+}
+
+fn format_time_elapsed(
+    time: Duration,
+    old_style: bool,
+) -> Result<String, chrono::format::ParseError> {
```
cakebaker, line 105: I'm not sure about this return type. The current implementation of the function never returns an `Err` and so the return type doesn't have to be a `Result`. On the other hand, it might make sense to use a `?` instead of `unwrap()` on line 106 and to adapt the error type accordingly.

src/uu/w/src/w.rs
```diff
@@ -283,6 +336,15 @@ mod tests {
         )
     }
 
+    #[test]
+    fn test_format_time_elapsed() {
+        let td = Duration::new(60 * 60 * 18 + 60 * 18, 0);
+        assert_eq!(
+            format_time_elapsed(td, false).unwrap(),
+            String::from("18:18m")
+        );
```
cakebaker, line 345: What's the expected output if you pass `true` to the `format_time_elapsed()` function?

src/uu/w/src/w.rs
aisuneko: This is to fix CI errors as mentioned earlier by @Krysztal112233

src/uu/w/src/w.rs
aisuneko: My intention was to comply with what we had earlier with `format_time`, which is also of this return type. Though in fact the `Result` on line 106 seems to have type of `chrono::OutOfRangeError`. I'll adapt that type instead.

Krysztal112233 (approved):
Great work, thanks

cakebaker:
Thanks for your PR!

PR #364, opened 2025-03-24
skill: add basic implementation of `skill`

maplestarplayl:
This pr add  the `skill` command, but only works on Linux


maplestarplayl:
Although the main functionality has been completed, some edge cases and potential improvements still need to be considered, such as command usage clarification, test coverage, more elegant implementation and performance optimization.

Krysztal112233:
Great work :)

src/uu/skill/src/command.rs
```diff
@@ -0,0 +1,105 @@
...
+            list: args.get_flag("list"),
+            table: args.get_flag("table"),
+            no_action: args.get_flag("no-action"),
+            verbose: args.get_flag("verbose"),
+            warnings: args.get_flag("warnings"),
+        }
+    }
+}
+
+pub fn parse_command(args: &mut impl uucore::Args) -> Vec<OsString> {
```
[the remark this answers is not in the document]
maplestarplayl: Ok, I can explain, i know it looks a little wired :)

# Issue Explanation:

The core problem lies in the signal format expected by skill, which follows the structure:
skill [signal] [options] <expression

However, the signal is passed in the form `-HUP` instead of `-s HUP`, preventing clap from automatically parsing it from the arguments.

# Solution Approach:
To handle this correctly:

- Signal as an Indexed Argument:
We must define signal as an indexed argument in clap so it can be properly parsed.
- Expression as an Indexed Argument
Since the `expression` is also passed in a similar way(not an option), it must also be treated as an indexed argument.

But after doing things above, another problem comes in: **Clap's Restriction on Optional Arguments**:
clap enforces a rule where if a later indexed argument is required, earlier ones cannot be optional.
This means `signal` must be marked as required to comply with clap's parsing logic, **which is conflict with skill's Default Behavior**:
The standard skill implementation (procps-ng 3.3.17) allows `signal` to be optional with a default value.
To reconcile this with clap's requirements, we preprocess the arguments before passing them to clap, ensuring signal is always provided (even if defaulted) to satisfy parsing constraints.

# Summary of the Issue

The root cause of these challenges stems from the fact that skill's argument format (as defined in procps-ng 3.3.17) cannot be directly modeled using clap's parsing rules.

However, this doesn’t imply any limitation in clap—rather, skill’s unconventional syntax (e.g., -HUP instead of -s HUP) deviates from typical CLI conventions, making parsing more complex.

Whatever, I will add more comments to the code to make it more reasonable :)
Also, if there are other more elegant way to process that, please tell me, thanks!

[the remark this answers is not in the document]
Krysztal112233: @cakebaker can you give us some help? seems like this is something clap can't handle.

src/uu/skill/src/command.rs
```diff
@@ -0,0 +1,105 @@
+use clap::ArgMatches;
+use std::collections::HashSet;
+use std::ffi::OsString;
+
+pub const SIGNALS: &[&str] = &[
+    "HUP", "INT", "QUIT", "ILL", "TRAP", "ABRT", "BUS", "FPE", "KILL", "USR1", "SEGV", "USR2",
+    "PIPE", "ALRM", "TERM", "STKFLT", "CHLD", "CONT", "STOP", "TSTP", "TTIN", "TTOU", "URG",
+    "XCPU", "XFSZ", "VTALRM", "PROF", "WINCH", "POLL", "PWR", "SYS",
+];
+const DEFAULT_SIGNAL: &str = "TERM";
```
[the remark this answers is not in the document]
maplestarplayl: Ok, got it.
But I found that the defined ALL_SIGNALS have 32 signals in linux , which is not consistent with the Linux Manual (has 31).
Do you know what causes this difference and how should I handle this?
Thanks for help
image

src/uu/skill/src/command.rs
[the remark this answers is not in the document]
Krysztal112233: @maplestarplayl and if it is possible, could you write some document(comment) and unit tests to make sure it doesn't regressing in future?

src/uu/skill/src/command.rs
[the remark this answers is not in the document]
Krysztal112233: please use text (code blocks) instead of picture, using text will make it easier to be searched :)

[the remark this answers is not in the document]
Krysztal112233: this post can help you to understand the `EXIT` "signal"

https://unix.stackexchange.com/questions/385030/are-exit-debug-return-and-err-signals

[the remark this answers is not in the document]
maplestarplayl: > please use text (code blocks) instead of picture, using text will make it easier to be searched :)

Sorry, my mistake. I’ll stick to using text in our future conversations.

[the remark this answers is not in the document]
maplestarplayl: > this post can help you to understand the `EXIT` "signal", which can be considered as a trap to capture something. it's an extension of shell, bash, zsh, fish, etc..
> 
> https://unix.stackexchange.com/questions/385030/are-exit-debug-return-and-err-signals
> 
> https://fishshell.com/docs/current/cmds/trap.html

After reviewing these articles, it appears that signals like `EXIT` are specific to the shell and not applicable to all processes. Therefore, in the context of the skill command, I have excluded them. Does this approach seem correct?

src/uu/skill/src/command.rs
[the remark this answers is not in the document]
maplestarplayl: > @maplestarplayl and if it is possible, could you write some document(comment) and unit tests to make sure it doesn't regressing in future?

Have added corresponding tests in the `command.rs` file

src/uu/skill/src/util.rs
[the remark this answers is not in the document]
maplestarplayl: Thanks, I’ll take a look at it.

src/uu/skill/src/command.rs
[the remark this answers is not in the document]
Krysztal112233: I think it can be processed in this way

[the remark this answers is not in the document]
Krysztal112233: > Sorry, my mistake. I’ll stick to using text in our future 

Please don't be nervous, it's not a mistake :)

src/uu/skill/src/command.rs
```diff
@@ -0,0 +1,161 @@
...
+            no_action: args.get_flag("no-action"),
+            verbose: args.get_flag("verbose"),
+            warnings: args.get_flag("warnings"),
+        }
+    }
+}
+
+// Pre-parses the command line arguments and returns a vector of OsString
+// Mainly used to parse the signal to make sure it is valid
+// and insert the default signal if it's not present
```
Krysztal112233, line 73:
```suggestion
// Pre-parses the command line arguments and returns a vector of OsString
//
// Mainly used to parse the signal to make sure it is valid
// and insert the default signal if it's not present
```

src/uu/skill/src/command.rs
```diff
@@ -0,0 +1,161 @@
...
+    // Flags
+    pub fast: bool,
+    pub interactive: bool,
+    pub list: bool,
+    pub table: bool,
+    pub no_action: bool,
+    pub verbose: bool,
+    pub warnings: bool,
+}
+
```
Krysztal112233, line 24: Maybe called `Settings` would be better?

maplestarplayl: Have changed it. Thanks for the suggestion!

maplestarplayl:
Refactor the util functions with pgrep's process.rs, and add corresponding tests for command and user usage in `skill` command

src/uu/skill/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+clap = { workspace = true }
+nix = { workspace = true }
+uu_pgrep = { path = "../pgrep" }
+
+[lib]
+path = "src/skill.rs"
+
+[[bin]]
+name = "skill"
+path = "src/main.rs"
```
cakebaker, line 25: Github shows that there is a newline missing at the end of this file.

cakebaker:
Can you please run `cargo clippy`? And the CI shows some issues on macOS :|

tests/by-util/test_skill.rs
```diff
@@ -0,0 +1,152 @@
...
+
+#[cfg(target_os = "linux")]
+#[test]
+fn test_missing_expression() {
+    new_ucmd!().fails().no_stdout().code_is(1).stderr_contains(
+        "the following required arguments were not provided:
+  <expression>...",
+    );
+}
+#[cfg(target_os = "linux")]
```
cakebaker, line 17:
```suggestion
}

#[cfg(target_os = "linux")]
```

tests/by-util/test_skill.rs
```diff
@@ -0,0 +1,152 @@
...
+#[test]
+fn test_invalid_signal() {
+    new_ucmd!()
+        .arg("-INVALID")
+        .fails()
+        .code_is(2)
+        .no_stdout()
+        .stderr_contains("Invalid signal: INVALID");
+}
+#[cfg(target_os = "linux")]
```
cakebaker, line 37:
```suggestion
}

#[cfg(target_os = "linux")]
```

tests/by-util/test_skill.rs
```diff
@@ -0,0 +1,152 @@
...
+#[test]
+fn test_mutiple_signal() {
```
cakebaker, line 39: A typo:
```suggestion
fn test_multiple_signal() {
```

tests/by-util/test_skill.rs
```diff
@@ -0,0 +1,152 @@
...
+    new_ucmd!()
+        .arg("-HUP")
+        .arg("-TERM")
+        .fails()
+        .code_is(2)
+        .no_stdout()
+        .stderr_contains("Too many signals");
+}
+#[cfg(target_os = "linux")]
```
cakebaker, line 48:
```suggestion
}

#[cfg(target_os = "linux")]
```

tests/by-util/test_skill.rs
```diff
@@ -0,0 +1,152 @@
...
+fn test_verbose_option() {
+    new_ucmd!()
+        .arg("-n") // no action
+        .arg("-v")
+        .arg("1")
+        .succeeds()
+        .stdout_contains("Would send signal TERM to process 1")
+        .no_stderr();
+}
+#[cfg(target_os = "linux")]
```
cakebaker, line 59:
```suggestion
}

#[cfg(target_os = "linux")]
```

tests/by-util/test_skill.rs
```diff
@@ -0,0 +1,152 @@
...
+
+#[cfg(target_os = "linux")]
+#[test]
+fn test_list_option_long() {
+    new_ucmd!()
+        .arg("--list")
+        .succeeds()
+        .no_stderr()
+        .stdout_contains("HUP INT QUIT ILL TRAP ABRT BUS FPE KILL USR1 SEGV USR2 PIPE ALRM TERM STKFLT CHLD CONT STOP TSTP TTIN TTOU URG XCPU XFSZ VTALRM PROF WINCH POLL PWR SYS");
+}
```
cakebaker, line 77: Here it might make sense to combine the functions into one and simply loop through the args, something like:
```rust
for arg in ["-l", "--list"] {
    new_ucmd!()
        .arg(arg)
...
}
```

tests/by-util/test_skill.rs
```diff
@@ -0,0 +1,152 @@
...
+    new_ucmd!()
+        .arg("--table")
+        .succeeds()
+        .no_stderr()
+        .stdout_contains("1 HUP     2 INT     3 QUIT    4 ILL     5 TRAP    6 ABRT    7 BUS")
+        .stdout_contains("8 FPE     9 KILL   10 USR1   11 SEGV   12 USR2   13 PIPE   14 ALRM")
+        .stdout_contains("15 TERM   16 STKFLT 17 CHLD   18 CONT   19 STOP   20 TSTP   21 TTIN")
+        .stdout_contains("22 TTOU   23 URG    24 XCPU   25 XFSZ   26 VTALRM 27 PROF   28 WINCH")
+        .stdout_contains("29 POLL   30 PWR    31 SYS");
+}
```
cakebaker, line 105: Same with these two functions.

tests/by-util/test_skill.rs
```diff
@@ -0,0 +1,152 @@
...
+
+#[cfg(target_os = "linux")]
+#[test]
+fn test_mutiple_options() {
+    new_ucmd!()
+        .arg("-nv") // no action + verbose
+        .arg("1")
+        .succeeds()
+        .stdout_contains("Would send signal TERM to process 1");
+}
```
cakebaker, line 115: This function is identical to `test_default_signal`, so it can be removed.
```suggestion
```

src/uu/skill/src/util.rs
```diff
@@ -0,0 +1,139 @@
+// This file is part of the uutils procps package.
+//
+// For the full copyright and license information, please view the LICENSE
+// file that was distributed with this source code.
+
+use std::collections::HashSet;
+use uu_pgrep::process::{walk_process, ProcessInformation};
+#[cfg(target_os = "linux")]
```
cakebaker, line 8:
```suggestion
use uu_pgrep::process::{walk_process, ProcessInformation};

#[cfg(target_os = "linux")]
```

src/uu/skill/src/util.rs
```diff
@@ -0,0 +1,139 @@
...
+            }
+        }
+    }
+
+    matching_processes
+}
+
+#[cfg(target_os = "linux")]
+pub fn get_process_owner(process: &mut ProcessInformation) -> Option<String> {
+    // let status_path = format!("/proc/{}/status", pid);
```
cakebaker, line 73: I would either remove this line or add a TODO with a reason why it's commented out.

src/uu/skill/src/command.rs
```diff
@@ -0,0 +1,162 @@
...
+            warnings: args.get_flag("warnings"),
+        }
+    }
+}
+
+// Pre-parses the command line arguments and returns a vector of OsString
+
+// Mainly used to parse the signal to make sure it is valid
+// and insert the default signal if it's not present
+pub fn parse_command(args: &mut impl uucore::Args) -> Vec<String> {
```
cakebaker, line 75: As this function can fail, I would return a `Result`.

src/uu/skill/src/command.rs
```diff
@@ -0,0 +1,162 @@
...
+        .filter(|arg: &&String| !exprs(arg))
+        .filter(|arg: &&String| !options(arg))
+        .collect::<Vec<&String>>();
+    if signals.len() == 1 {
+        let signal = &signals[0].as_str()[1..]; // Remove the leading '-'
+        if ALL_SIGNALS.contains(&signal) {
+            args.to_vec()
+        } else {
+            eprintln!("Invalid signal: {}", &signal);
+            std::process::exit(2);
```
cakebaker, line 115: See my previous comment about returning a `Result`. We usually use something like:
```rust
return Err(USimpleError::new(2, format!("Invalid signal: {}", &signal)));
```

src/uu/skill/src/command.rs
```diff
@@ -0,0 +1,162 @@
...
+        }
+    } else if signals.is_empty() {
+        // If no signal is provided, return the original args with default signal
+        let mut new = args.to_vec();
+        new.insert(1, ALL_SIGNALS[DEFAULT_SIGNAL].to_string());
+        new
+    } else {
+        eprintln!("Too many signals");
+        std::process::exit(2);
```
cakebaker, line 124: See my previous comment.

cakebaker (commented):
Overall good job :)

PR #371, opened 2025-03-27
slabtop: implemented tui

Krysztal112233:


cakebaker:
Yep, there is definitely some performance problem, it constantly uses ~70% of CPU on my machine ;-)

Krysztal112233:
CPU usage is now reduced by slowing down the redraw speed, but redrawing is still triggered once the window size is changed.

But I think it's as expected, as it avoids confusing TUI output :)

Krysztal112233:
Hi, this pr has been finished for a long time. I've reorganized the commits and could you have a look about it? :)

@cakebaker @sylvestre

cakebaker:
It looks like you added some unrelated stuff with your latest commit :|

PR #378, opened 2025-03-29
pmap: fixed to hide the header when permission is denied

### LINKED ISSUE #376: pmap: don't output header if `/proc/<PID>/smaps` can't be accessed
While reviewing https://github.com/uutils/procps/pull/368 I noticed that we output a header if the permission is denied to read `/proc/<PID>/smaps` whereas the original `pmap` doesn't show a header. On my machine, this seems only to affect PID 1.
```
$ cargo run -q pmap -x 1
1:   /sbin/init \vmlinuz-linux
Address           Kbytes     RSS   Dirty Mode  Mapping
$ pmap -x 1
1:   /sbin/init \vmlinuz-linux
```

estodi:
Resolves #376 

As the issue is not limited to `smaps`, several parts of the code are modified.

### DIFF
--- a/src/uu/pmap/src/pmap.rs
+++ b/src/uu/pmap/src/pmap.rs
@@ -81,12 +81,16 @@ fn parse_cmdline(pid: &str) -> Result<String, Error> {
     Ok(cmdline.into())
 }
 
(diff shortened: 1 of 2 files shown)


cakebaker:
Thanks!

PR #401, opened 2025-04-25
top, w: fix uptime format

Bluemangoo:
procps uptime format from procps differs from coreutils'. 

for example:

coreutils:

```text
up 00:42
```

procps v4.0.5:

```text
up 42 min
```

also see [coreutils/7757](https://github.com/uutils/coreutils/pull/7757)

### DIFF
--- a/src/uu/top/src/header.rs
+++ b/src/uu/top/src/header.rs
@@ -5,9 +5,8 @@
 
 use crate::picker::sysinfo;
 use bytesize::ByteSize;
-use uucore::uptime::{
-    get_formated_uptime, get_formatted_loadavg, get_formatted_nusers, get_formatted_time,
-};
+use uu_w::get_formatted_uptime_procps;
+use uucore::uptime::{get_formatted_loadavg, get_formatted_nusers, get_formatted_time};
 
 pub(crate) fn header(scale_summary_mem: Option<&String>) -> String {
     format!(
@@ -39,8 +38,9 @@ fn format_memory(memory_b: u64, unit: u64) -> f64 {
     ByteSize::b(memory_b).0 as f64 / unit as f64
 }
 
+#[inline]
 fn uptime() -> String {
-    get_formated_uptime(None).unwrap_or_default()
+    get_formatted_uptime_procps().unwrap_or_default()
 }
 
 #[cfg(target_os = "linux")]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1655,6 +1655,7 @@ dependencies = [
  "prettytable-rs",
  "sysinfo",
  "uu_vmstat",
+ "uu_w",
  "uucore",
  "windows-sys 0.59.0",
 ]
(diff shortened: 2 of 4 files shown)


src/uu/top/src/header.rs
```diff
@@ -39,12 +40,39 @@ fn format_memory(memory_b: u64, unit: u64) -> f64 {
     ByteSize::b(memory_b).0 as f64 / unit as f64
 }
 
+fn format_uptime_procps(up_secs: i64) -> UResult<String> {
```
cakebaker, line 43: What's the reason for using a `procps` suffix in the function name (and in `get_formated_uptime_procps`)?

src/uu/top/src/header.rs
```diff
@@ -39,12 +40,39 @@ fn format_memory(memory_b: u64, unit: u64) -> f64 {
...
+    let hour_min_str = if up_hours > 0 {
+        format!("{up_hours:2}:{up_mins:02}")
+    } else {
+        format!("{up_mins:02} min")
+    };
+    Ok(format!("{}{}", day_str, hour_min_str))
+}
+
+#[inline]
+fn get_formated_uptime_procps() -> UResult<String> {
```
cakebaker, line 64: A typo:
```suggestion
fn get_formatted_uptime_procps() -> UResult<String> {
```

src/uu/top/src/header.rs
Bluemangoo: to distinguish it from that of uucore

Bluemangoo: should I keep it?

cakebaker: Yes, I would keep it. Currently I don't have a better idea. And we can still change it later, if necessary.

src/uu/top/src/header.rs
```diff
@@ -249,3 +277,31 @@ fn memory(scale_summary_mem: Option<&String>) -> String {
...
+            60,
+            100,
+            200,
+            3600,
+            5000,
+            3600 * 24,
+            3600 * 24 + 60,
+            3600 * 24 + 3600,
+        ];
+        let re = regex::Regex::new(r"(\d+ days?,)?\s*\d+( min|:\d+)").unwrap();
```
cakebaker, line 299: I wouldn't use a regex for this test as it also matches invalid results. I think it is simpler to use `(input, expected)` pairs.

Bluemangoo:
Are we planning a common crate for uutils/procps, which similar to uucore?

I think the uptime formatter are widely used in procps. As the alternative option, I would place it in `w`, since `top` requires `systemd-dev` on Linux.

cakebaker:
> Are we planning a common crate for uutils/procps, which is similar to uucore?

Yes, I think it would make sense to have such a crate.

Bluemangoo:
all fixed.

cakebaker:
Thanks!

PR #418, opened 2025-05-15
pmap: implemented show-path option

estodi:
Added support for the show-path option.

### DIFF
--- a/tests/by-util/test_pmap.rs
+++ b/tests/by-util/test_pmap.rs
@@ -30,7 +30,7 @@ fn test_existing_pid() {
         .succeeds()
         .stdout_move_str();
 
-    assert_format(pid, &result);
+    assert_format(pid, &result, false);
 }
 
 #[test]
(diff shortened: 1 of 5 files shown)


tests/by-util/test_pmap.rs
```diff
@@ -211,14 +268,18 @@ fn assert_cmdline_only(pid: &str, s: &str) {
...
     assert!(re.is_match(first_line));
 
     let rest = rest.trim_end();
     let (memory_map, last_line) = rest.rsplit_once('\n').unwrap();
-    let re = Regex::new(r"(?m)^[0-9a-f]{16} +[1-9][0-9]*K (-|r)(-|w)(-|x)(-|s)- (  \[ (anon|stack) \]|[a-zA-Z0-9._-]+)$").unwrap();
+    let re = if show_path {
+        Regex::new(r"(?m)^[0-9a-f]{16} +[1-9][0-9]*K (-|r)(-|w)(-|x)(-|s)- (  \[ (anon|stack) \]|/[/a-zA-Z0-9._-]+)$").unwrap()
+    } else {
+        Regex::new(r"(?m)^[0-9a-f]{16} +[1-9][0-9]*K (-|r)(-|w)(-|x)(-|s)- (  \[ (anon|stack) \]|[a-zA-Z0-9._-]+)$").unwrap()
+    };
```
cakebaker, line 282: Maybe it would make sense to split the regexes to make it easier to see the (small) differences? Something like:
```rust
let base_pattern = "<pattern for the identical part>";
let mapping_pattern = if show_path {
    "<pattern path expected>"
} else {
    "<pattern no path expected>"
}

let re = Regex::new(&format("{base_pattern}{mapping_pattern}")).unwrap();
```

cakebaker:
Good work, thanks!

PR #421, opened 2025-05-17
pgrep/pkill/pidwait: Implement --logpidfile (sic)

dezgeg:


tests/by-util/test_pgrep.rs
```diff
@@ -601,5 +601,71 @@ fn test_pidfile_nonexistent_file() {
...
+        .succeeds()
+        .stdout_is("1\n");
+
+    flock_process.kill().unwrap();
+    flock_process.wait().unwrap();
+}
+
+#[test]
+#[cfg(target_os = "linux")]
+#[ignore = "flock --fcntl not supported on CI runner"]
```
cakebaker, line 648: Our test framework provides an `is_ci()` function you could use to make the test run locally.

dezgeg: Ah, that's better. changed

src/uu/pgrep/src/process_matcher.rs
```diff
@@ -411,8 +414,48 @@ fn test_parse_pidfile_content_valid() {
     assert_eq!(parse_pidfile_content("\n123\n"), None);
 }
 
-pub fn read_pidfile(filename: &str) -> UResult<i64> {
-    let content = fs::read_to_string(filename)
+#[cfg(unix)]
+fn check_pidfile_locked(file: &std::fs::File) -> bool {
```
cakebaker, line 418: A detail: I would rename the function to `is_pidfile_locked` (or `is_locked`). Otherwise the condition on line 451, `if check_locked && !check_pidfile_locked(&file)`, reads a bit awkward: `check_locked` is about whether a lock check should be performed whereas `check_pidfile_locked` is about doing the lock check (and not about whether a different lock check should be performed).

dezgeg: Indeed, I had it originally return UResult<()> so `check_` made more sense then but not for the predicate version. Changed to `is_`

cakebaker:
Thanks!

PR #423, opened 2025-05-20
pmap: implemented quiet option

estodi:
Added support for the quiet option.

### DIFF
--- a/tests/by-util/test_pmap.rs
+++ b/tests/by-util/test_pmap.rs
@@ -30,7 +30,7 @@ fn test_existing_pid() {
         .succeeds()
         .stdout_move_str();
 
-    assert_format(pid, &result, false);
+    assert_format(pid, &result, false, false);
 }
 
 #[test]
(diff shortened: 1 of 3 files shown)


tests/by-util/test_pmap.rs
```diff
@@ -284,8 +309,10 @@ fn assert_format(pid: u32, s: &str, show_path: bool) {
     let re = Regex::new(&format!("{base_pattern}{mapping_pattern}")).unwrap();
     assert!(re.is_match(memory_map));
 
-    let re = Regex::new("^ total +[1-9][0-9]*K$").unwrap();
-    assert!(re.is_match(last_line));
+    if !quiet {
+        let re = Regex::new("^ total +[1-9][0-9]*K$").unwrap();
+        assert!(re.is_match(last_line));
+    }
```
cakebaker, line 315: You have to handle both the `if` and the `else` case. Otherwise, if `quiet` is `true`, you skip the assert without ensuring there is no "total" line. And thus the test passes even without the changes you made to `pmap.rs` and `pmap_config.rs` ;-)

cakebaker:
Thanks!

PR #429, opened 2025-05-31
vmstat: implement `--timestamp`

Bluemangoo:


cakebaker (commented):
The changes look good. Can you please add a test to ensure we don't regress in the future?

PR #452, opened 2025-06-27
w: implement `--container`

Bluemangoo:


Bluemangoo:
clippy complains `error: variables can be used directly in the `format!` string`. caused by the update([action](https://github.com/uutils/procps/actions/runs/15918050247/job/44899267648)):

```
stable-x86_64-unknown-linux-gnu updated - rustc 1.88.0 (6b00bc388 2025-06-23) (from rustc 1.87.0 (17067e9ac 2025-05-09))
```

should we ignore(allow) it? or I can fix them in another pr (almost all crates are involved).

Krysztal112233:
I'll fix it later

Bluemangoo:
fixed.

PR #457, opened 2025-07-05
pkill: implement `--queue`

Bluemangoo:
wrote a tool to test:

```
$ ./sigr                    
I'm 99648
recv a sig = 2 data = 10 data = 10
```

```
$ cargo run -q pkill -SIGINT -q 10 sigr -e
./sigr killed (pid 99648)
```

### DIFF
--- a/src/uu/pkill/src/pkill.rs
+++ b/src/uu/pkill/src/pkill.rs
@@ -4,7 +4,7 @@
 // file that was distributed with this source code.
 
 // Pid utils
-use clap::{arg, crate_version, Command};
+use clap::{arg, crate_version, value_parser, Command};
 #[cfg(unix)]
 use nix::{
     sys::signal::{self, Signal},
@@ -48,7 +48,7 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
     } else {
         let sig = (settings.signal as i32)
             .try_into()
-            .map_err(|e| std::io::Error::from_raw_os_error(e as i32))?;
+            .map_err(|e| Error::from_raw_os_error(e as i32))?;
         Some(sig)
     };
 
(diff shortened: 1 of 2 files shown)


Bluemangoo:
Common distributions do not have commands to detect output queues. The tool mentioned above is written in C++.

It might be difficult to test if it really working only with rust tests. btw, `test_pkill.rs` only contains tests for invalid arguments. I think we can open an issue for this.

```c++
#include <string.h>
#include <signal.h>
#include <unistd.h>
#include <stdlib.h>
#include <stdio.h>
 
void handler(int, siginfo_t *, void*);
 
int main(int argc, char *argv[])
{
    printf("I'm %d\n", getpid());
    struct sigaction act;
    act.sa_sigaction = handler;
    sigemptyset(&act.sa_mask);
    act.sa_flags = SA_SIGINFO;
    
    if(sigaction(SIGINT, &act, NULL) < 0) {
        perror("sigaction error");
        exit(0);
    }
    for(; ;)
      pause();
    return 0;
}
 
void handler(int sig, siginfo_t *info, void *ctx)
{
    printf("recv a sig = %d data = %d data = %d\n", sig, info->si_value.sival_int, info->si_int);
}
```

src/uu/pkill/src/pkill.rs
```diff
@@ -71,25 +73,53 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
...
-        if let Err(e) = signal::kill(Pid::from_raw(pid.pid as i32), sig) {
+        #[cfg(target_os = "linux")]
+        let result = if let Some(queue) = queue {
+            let v = unsafe {
+                sigqueue(
+                    pid.pid as i32,
+                    match sig {
+                        Some(s) => s as uucore::libc::c_int,
+                        None => 0,
+                    },
```
cakebaker, line 110: I think you can simplify this `match` with `map_or`:
```suggestion
                    sig.map_or(0, |s| s as uucore::libc::c_int),
```

Bluemangoo:
fixed.

cakebaker:
Thanks!

PR #466, opened 2025-07-14
skill: add basic implementation

Bluemangoo:


src/uu/skill/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+[package]
+name = "uu_skill"
+description = "snice - (uutils) send a signal or report process status"
+repository = "https://github.com/uutils/procps/tree/main/src/uu/snice"
```
cakebaker, line 4:
```suggestion
repository = "https://github.com/uutils/procps/tree/main/src/uu/skill"
```

src/uu/snice/src/snice.rs
```diff
@@ -66,86 +69,25 @@ impl SignalDisplay {
...
-        };
-
-        let display = if matches.get_flag("table") {
-            Some(SignalDisplay::Table)
-        } else if matches.get_flag("list") {
-            Some(SignalDisplay::List)
-        } else {
-            None
-        };
+pub fn print_signals(settings: &Settings) -> Option<()> {
```
cakebaker, line 72: The return type is unexpected. I think it is not necessary if you make the function do what its name says: to print signals. And to move the decision, whether the user wants to print the signals, to the caller of the function. Something like:
```rust
// in uumain
if let Some(display) = settings.display {
    print_signals(display);
    return Ok(());
}
```

src/uu/skill/src/skill.rs
```diff
@@ -0,0 +1,114 @@
...
+            let output = construct_verbose_result(&pids, &results).trim().to_owned();
+            println!("{output}");
+        }
+    }
+
+    Ok(())
+}
+
+#[cfg(unix)]
+fn perform_action(pids: &[u32], signal: &Signal) -> Vec<Option<ActionResult>> {
```
cakebaker, line 66: I think the `Option` in the return type is not necessary and you can just use `Vec<ActionResult>`.

src/uu/skill/src/skill.rs
```diff
@@ -0,0 +1,114 @@
...
+                    Ok(_) => ActionResult::Success,
+
+                    Err(_) => ActionResult::PermissionDenied,
+                })
+            }
+            #[cfg(not(unix))]
+            {
+                // On non-unix platforms, we cannot send signals
+                None
+            }
```
cakebaker, line 81: This block is unreachable because of the `#[cfg(unix)]` on line 65.

src/uu/snice/src/snice.rs
```diff
@@ -155,27 +97,26 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
...
-            return Ok(());
-        }
+    if print_signals(&settings).is_some() {
+        return Ok(());
     }
 
     // Case1: Perform priority
     if let Some(targets) = settings.expressions {
+        let priority_str = matches
+            .try_get_one::<String>("priority")
```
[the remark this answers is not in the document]
Bluemangoo: I removed `priority` from `Settings`, the code is moved from there:

https://github.com/uutils/procps/blob/1be5929a248238db51d5a77eb5c35ed01ff375be/src/uu/snice/src/snice.rs#L79-L82

I have no idea about it. Should I change it to `get_one`?

Bluemangoo:
> Sorry, there is a conflict due to your other PR :|

fixed.

PR #479, opened 2025-08-07
top: implement `PR` `NI`

Bluemangoo:
The old PR impl is exactly NI impl

### DIFF
--- a/src/uu/top/src/picker.rs
+++ b/src/uu/top/src/picker.rs
@@ -26,6 +26,7 @@ pub(crate) fn pickers(fields: &[String]) -> Vec<Box<dyn Fn(u32) -> String>> {
             "PID" => helper(pid),
             "USER" => helper(user),
             "PR" => helper(pr),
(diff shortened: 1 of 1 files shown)


src/uu/top/src/picker.rs
```diff
@@ -76,11 +77,42 @@ fn user(pid: u32) -> String {
...
+    // normal processes
+    if policy == SCHED_OTHER || policy == SCHED_BATCH || policy == SCHED_IDLE {
+        return (get_nice(pid) + 20).to_string();
+    }
+
+    // real-time processes
+    let param: *mut sched_param = unsafe { malloc(size_of::<sched_param>()) as *mut sched_param };
+    unsafe { sched_getparam(pid as c_int, param) };
+    let priority = unsafe { (*param).sched_priority };
+    unsafe { free(param as *mut c_void) };
```
cakebaker, line 97: What's the reason for the manual memory management?

PR #483, opened 2025-08-10
skill: implement `no-action`

Bluemangoo:


Bluemangoo:
close as it is included in another pr

PR #484, opened 2025-08-10
skill&snice: implement `no-action` `warnings` `interactive`

Bluemangoo:
- impl `no-action` for skill
- impl `warnings` for skill&snice
- impl `interactive` for skill&snice


tests/by-util/test_snice.rs
```diff
@@ -16,3 +16,9 @@ fn test_no_args() {
 fn test_no_process_selected() {
     new_ucmd!().arg("-u=invalid_user").fails().code_is(1);
 }
+
+#[test]
+fn test_interrupt_args() {
+    new_ucmd!().args(&["-i", "-v"]).fails().code_is(1);
+    new_ucmd!().args(&["-i", "-n"]).fails().code_is(1);
+}
```
Krysztal112233, line 24:
```suggestion
#[test]
fn test_interactive_conflict_args() {
    new_ucmd!().args(&["-i", "-v"]).fails().code_is(1);
    new_ucmd!().args(&["-i", "-n"]).fails().code_is(1);
}
```

I think rename it into `test_interactive_conflict_args` or other name would be better?

Bluemangoo: fixed.

PR #492, opened 2025-08-11
ps: Implement bunch of field selection flags

dezgeg:


src/uu/ps/src/mapping.rs
```diff
@@ -34,9 +93,16 @@ pub(crate) fn default_mapping() -> HashMap<String, String> {
     // Those mapping generated from manpage
     append("%cpu", "%CPU");
     append("%mem", "%MEM");
+    append("_left", "LLLLLLLL");
+    append("_left2", "L2L2L2L2");
+    append("_right", "RRRRRRRR");
+    append("_right2", "R2R2R2R2");
+    append("_unlimited", "U");
+    append("_unlimited2", "U2");
```
Krysztal112233, line 101: Could you add document to explain where those mappings comes from? I don't find then in the manpage of `ps`

dezgeg: I updated the comment. This commit just adds `ps L` ones but there are also undocumented ones (that are internally used by some of the built-in field choices)

Krysztal112233:
Awesome work!

PR #506, opened 2025-08-26
top: implement tui

Bluemangoo:


Krysztal112233:
wow, this TUI is really beautiful.

but humm..



I think I have too many cores to display.

Krysztal112233 (approved):

PR #509, opened 2025-09-04
top: tui implement `4`, `l`

Bluemangoo:
implement: 

- `4`: set columns for cpu&memory graph in header
- `l`: hide or show load average line (the first line)

resolves the [issue](https://github.com/uutils/procps/pull/506#issuecomment-3248489220) that @Krysztal112233 mentioned.

### DIFF
--- a/src/uu/top/src/tui/mod.rs
+++ b/src/uu/top/src/tui/mod.rs
@@ -293,15 +293,27 @@ impl<'a> Tui<'a> {
                     bar_content,
                     false,
                 ));
-                mem_bars.push((
-                    format!("{unit_name} Swap"),
-                    mem.used_swap as f64 / mem.total_swap as f64 * 100.0,
-                    format_memory(mem.total_swap, unit),
-                    0.0,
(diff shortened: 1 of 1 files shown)


Krysztal112233 (approved):

src/uu/top/src/tui/mod.rs
```diff
@@ -34,50 +34,114 @@ impl<'a> Tui<'a> {
...
 
     fn calc_header_height(&self) -> u16 {
-        let mut height = 1;
+        let mut height = 0;
 
-        if self.stat.cpu_graph_mode != CpuGraphMode::Hide {
+        if self.stat.show_load_avg {
             height += 1;
-            height += self.header.cpu.len() as u16;
+        }
```
cakebaker, line 41: I think you can simplify the initialization of `height` by using `from`:
```rust
let mut height = u16::from(self.stat.show_load_avg);
```

cakebaker:
Thanks!

PR #518, opened 2025-09-08
top: implement tui `2` `3` `c` `z`

Bluemangoo:


src/uu/top/src/tui/color.rs
```diff
@@ -0,0 +1,60 @@
+use ratatui::prelude::Stylize;
```
cakebaker, line 1: Can you please add the usual file header?

src/uu/top/src/tui/input.rs
```diff
@@ -0,0 +1,222 @@
...
+    NumaNode,
+}
+
+pub fn handle_input(
+    e: Event,
+    settings: &Settings,
+    tui_stat: &RwLock<TuiStat>,
+    data: &RwLock<(Header, ProcList)>,
+    should_update: &AtomicBool,
+) -> UResult<bool> {
```
cakebaker, line 31: I think the return value could be just a `bool`.

PR #530, opened 2025-09-27
top: tui impl `u` `U` (filter by user)

Bluemangoo:


src/uu/top/src/tui/input.rs
```diff
@@ -315,5 +335,31 @@ fn handle_input_value(
             data.write().unwrap().0.update_cpu(&stat);
             should_update.store(true, Ordering::Relaxed);
         }
+        InputEvent::FilterUser | InputEvent::FilterEUser => {
+            let input_value = { tui_stat.read().unwrap().input_value.clone() };
+            let user = match try_into_uid(&input_value) {
```
cakebaker, line 340: Before passing `input_value` to `try_into_uid` you probably have to check for an empty value to reset the current filter.

Bluemangoo:
Sorry. Fixed now.

PR #549, opened 2025-10-18
top: impl `^u` `r` `k`

Bluemangoo:


src/uu/top/src/picker.rs
```diff
@@ -432,6 +432,33 @@ fn mem(pid: u32, _stat: Stat) -> Box<dyn Column> {
     )
 }
 
+#[cfg(target_os = "linux")]
+pub(crate) fn get_supplementary_groups(pid: u32) -> String {
+    use sysinfo::{Gid, Groups};
+    let groups = Groups::new_with_refreshed_list();
```
cakebaker:
```suggestion
    use sysinfo::{Gid, Groups};

    let groups = Groups::new_with_refreshed_list();
```

src/uu/top/src/picker.rs
```diff
@@ -432,6 +432,33 @@ fn mem(pid: u32, _stat: Stat) -> Box<dyn Column> {
...
+                    .map(|group| group.name())
+                    .collect::<Vec<_>>()
+                    .join(",");
+                return groups;
+            }
+        }
+        String::new()
+    } else {
+        String::new()
+    }
```
cakebaker: You can move the string creation outside of the `if let` block:
```suggestion
    }
    String::new()
```

cakebaker:
Good work, thanks!

PR #625, opened 2026-02-09
ps: remove libc in ps

Franklin-Qi:
Handling unsafe uucore::libc::getsid functions with Rustix

Closes: #597


Krysztal112233 (changes requested):
LGTM, but something can be better :)

PR #666, opened 2026-04-08
`w`: Fix formatting of command line value

alxndrv:
The command line value exposed by the kernel in `/proc/<pid>/cmdline` represents whitespace as NULL bytes.
Using `read_to_string()` on this directly will cause the output to join all parts of the original command into one long string.
This PR adds a very simple fix to avoid this.

### Before:
```bash
$ cargo run -q w
 22:16:22 up  2:00,  1 user,  load average: 0.32, 0.37, 0.38
USER     TTY       LOGIN@   IDLE   JCPU   PCPU  WHAT
alx      tty1      20:16    1:59m  0.00   0     /bin/sh/usr/bin/niri-session-l
```
### After:
```bash
$ cargo run -q w
 22:17:02 up  2:00,  1 user,  load average: 0.67, 0.46, 0.41
USER     TTY       LOGIN@   IDLE   JCPU   PCPU  WHAT
alx      tty1      20:16    2:00m  0.00   0     /bin/sh /usr/bin/niri-session -l 
```

### DIFF
--- a/src/uu/w/src/w.rs
+++ b/src/uu/w/src/w.rs
@@ -142,7 +142,8 @@ fn format_time(time: String) -> Result<String, chrono::format::ParseError> {
 #[cfg(target_os = "linux")]
 fn fetch_cmdline(pid: i32) -> Result<String, std::io::Error> {
     let cmdline_path = Path::new("/proc").join(pid.to_string()).join("cmdline");
-    fs::read_to_string(cmdline_path)
+    // See `proc_pid_cmdline(5)` man page for the format of /proc/<pid>/cmdline
+    fs::read_to_string(cmdline_path).map(|s| s.trim_end_matches('\0').replace('\0', " "))
 }
 
 #[cfg(target_os = "linux")]
@@ -452,11 +453,8 @@ mod tests {
     fn test_fetch_cmdline() {
         // uucore's utmpx returns an i32, so we cast to that to mimic it.
         let pid = process::id() as i32;
-        let path = Path::new("/proc").join(pid.to_string()).join("cmdline");
-        assert_eq!(
-            fs::read_to_string(path).unwrap(),
-            fetch_cmdline(pid).unwrap()
-        );
+        let cmdline = std::env::args().collect::<Vec<String>>().join(" ");
+        assert_eq!(cmdline, fetch_cmdline(pid).unwrap());
     }
 
     #[test]


Bluemangoo:
Can you fix the tests?

alxndrv:
Tests are fixed now, but I'm a bit confused by the coverage check. Is there anything I can do to make it come up as successful?

Krysztal112233:
I feel the current test coverage is not very important, and it's already sufficient :)

src/uu/w/src/w.rs
```diff
@@ -452,11 +456,8 @@ mod tests {
     fn test_fetch_cmdline() {
         // uucore's utmpx returns an i32, so we cast to that to mimic it.
         let pid = process::id() as i32;
-        let path = Path::new("/proc").join(pid.to_string()).join("cmdline");
-        assert_eq!(
-            fs::read_to_string(path).unwrap(),
-            fetch_cmdline(pid).unwrap()
-        );
+        let cmdline = std::env::args().collect::<Vec<String>>().join(" ");
+        assert_eq!(cmdline, fetch_cmdline(pid));
```
Krysztal112233: Could you tell me why it needs to be changed to this?

src/uu/w/src/w.rs
```diff
@@ -140,9 +140,13 @@ fn format_time(time: String) -> Result<String, chrono::format::ParseError> {
...
 
 #[cfg(target_os = "linux")]
-fn fetch_cmdline(pid: i32) -> Result<String, std::io::Error> {
+fn fetch_cmdline(pid: i32) -> String {
     let cmdline_path = Path::new("/proc").join(pid.to_string()).join("cmdline");
     fs::read_to_string(cmdline_path)
+        .unwrap_or_default()
+        .split_terminator("\0")
+        .collect::<Vec<_>>()
+        .join(" ")
```
Krysztal112233: I think removing `Result` might not be a good idea

Krysztal112233: I think this would better fit Rust's style :P 

```suggestion
fn fetch_cmdline(pid: i32) -> Result<String, std::io::Error> {
    let cmdline_path = Path::new("/proc").join(pid.to_string()).join("cmdline");
    fs::read_to_string(cmdline_path).map(|it| it.trim_end_matches('\0').replace('\0', " "))
}
```

Krysztal112233:
Can you add a comment for the `fetch_cmdline` function? I noticed that the man page mentions this point after checking it, so you could reference the man page there:

https://man7.org/linux/man-pages/man5/proc_pid_cmdline.5.html

Krysztal112233:
Great work, I think this work can inspire us to consider similar issues elsewhere :)

src/uu/w/src/w.rs
alxndrv: The original implementation was basically a 1:1 copy of the code that's inside the `fetch_cmdline` function (both just build a Path from a PID and `read_to_string` it).
So previously, we had a test where both sides of the `assert_eq` are results of basically identical code. This kind of approach guarantees that things "work", but does not test for correctness. Both sides could be outputting the same incorrect result, but the test would pass. You can probably see how that's potentially problematic :sweat_smile: 

Since we're already using `std::process::id()` to figure out the PID of the process, I thought it would make more sense to compare the result of our `fetch_cmdline` with a "known correct" value which in this case would be `std::env::args()`. Hope my explanation makes sense :sweat_smile:

alxndrv:
@Krysztal112233 Thanks for the thorough review. My Rust is... rusty, so these pointers are very much needed.
@Bluemangoo Feel free to hold off on the review until I address all of these comments :slightly_smiling_face:

src/uu/w/src/w.rs
Krysztal112233: @sylvestre could you please take a review for this point?

Bluemangoo (approved):
LGTM

Krysztal112233 (approved):

Krysztal112233:
Thanks for your contribution!

PR #668, opened 2026-04-09
vmstat: fix meminfo unit

### LINKED ISSUE #667: Fix wrong unit of meminfo parser
`meminfo` shows data like these:

```text
MemTotal:       30524188 kB
MemFree:         3695720 kB
MemAvailable:   20690088 kB
...
```

In `bytesize::ByteSize::from_str`, they will be parsed as KB([source of Unit::from_str](https://github.com/bytesize-rs/bytesize/blob/1cacb79b9d845741b070f6124baddc7f76cb57f1/src/parse.rs#L207-L209), called in [ByteSize::from_str](https://github.com/bytesize-rs/bytesize/blob/1cacb79b9d845741b070f6124baddc7f76cb57f1/src/parse.rs#L17)).

```rust
impl str::FromStr for Unit {
    // ...
    fn from_str(unit: &str) -> Result<Self, Self::Err> {
        match () {
            // ...
            _ if unit.eq_ignore_ascii_case("k") | unit.eq_ignore_ascii_case("kb") => {
                Ok(Self::KiloByte)
            }
            // ...
        }
    }
}
```

Anyway, it is formatted in `KiB` instead of `KB`. Or rather, `kB` has always been `KiB`. But bytesize ignored cases(`eq_ignore_ascii_case`).

bytesize said they won't fix this: bytesize-rs/bytesize#157

See sources of Linux kernel [here](https://github.com/torvalds/linux/blob/7f87a5ea75f011d2c9bc8ac0167e5e2d1adb1594/fs/proc/meminfo.c#L28-L32)

```c
static void show_val_kb(struct seq_file *m, const char *s, unsigned long num)
{
	seq_put_decimal_ull_width(m, s, num << (PAGE_SHIFT - 10), 8);
	seq_write(m, " kB\n", 4);
}
```

Thanks for @Franklin-Qi to [mention](https://github.com/uutils/procps/issues/389#issuecomment-4096530728) it in #389

Bluemangoo:
- fix #667

Use a simple but not very graceful way to fix the issue.

See more in the issue.

### DIFF
--- a/src/uu/vmstat/src/parser.rs
+++ b/src/uu/vmstat/src/parser.rs
@@ -211,6 +211,16 @@ pub struct Meminfo {
     pub swap_total: bytesize::ByteSize,
     pub swap_free: bytesize::ByteSize,
 }
+
+#[cfg(target_os = "linux")]
+fn kb_to_kib(mut size: bytesize::ByteSize) -> bytesize::ByteSize {
+    // "kB" means KiB instead of KB
+    // Convert from 1000-based(parsed by bytesize from "kB" ended string) to 1024-based(which it actually is)
+    // See more at https://github.com/uutils/procps/issues/667
+    size.0 = size.0 / 1000 * 1024;
+    size
+}
+
 #[cfg(target_os = "linux")]
 impl Meminfo {
     pub fn current() -> Self {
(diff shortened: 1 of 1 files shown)


Franklin-Qi:
LGTM :100:

robjtede (approved):
given the output of meminfo is non-standard, this fix lgtm

Krysztal112233:
oops. historical echoes from the kernel XD.

@Bluemangoo could you please also add a documentation link to this issue? https://github.com/bytesize-rs/bytesize/issues/157

Krysztal112233 (approved):

PR #673, opened 2026-04-11
actions: add security audit workflow

xtqqczze:
Action will be executed periodically at midnight of each day and check if any new advisories appear for crate dependencies.

For each new advisory (including informal) an issue will be created.

### DIFF
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -1,0 +1,17 @@
+name: Security audit
+
+# spell-checker:ignore (misc) rustsec
+
+on:
+  schedule:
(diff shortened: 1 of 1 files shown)


Krysztal112233:
@sylvestre @cakebaker  Could you please check if this is necessary?

xtqqczze:
We already use the action at uutils/coreutils and others.

Bluemangoo:
downvote, `renovate` is enough.

xtqqczze:
@renovate doesn't update everything, see https://github.com/uutils/procps/pull/675

PR #701, opened 2026-05-17
use rustix process functions

xtqqczze:
We want to remove the `rustix::process` functions, so use rustix directly in this crate.

cc: @cakebaker


cakebaker:
Thanks!

PR #400, opened 2025-04-24
pidwait: optimize & cross-platform

FurryR:
Resolves #399 (partially).

Needs to be reviewed before merge as it contains untested code. However it seems that at least the implementation for Linux works fine.

PR #711, opened 2026-05-29
pidwait: replace busy-loop with `pidfd_open` and `poll` on Linux

Krysztal112233:
This PR pick the Linux implementation from PR #400 to replace the inefficient busy-loop polling (`/proc` every 50ms) with `pidfd_open` + `poll`, which blocks efficiently in the kernel with zero CPU usage.

Also added a simple test to verify the wait works correctly.

Thanks for @FurryR

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2425,10 +2425,11 @@ name = "uu_pidwait"
 version = "0.0.1"
 dependencies = [
  "clap",
- "nix 0.30.1",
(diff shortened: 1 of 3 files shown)


FurryR:
Nice to see the new implementation here. Ima like dropping my pull request
Will see if I can help.

Bluemangoo (approved):

PR #778, opened 2026-09-09
Bump `uucore` and `nix`

cakebaker:
This PR bumps `uucore` from `0.7.0` to `0.11.0` and `nix` from `0.30.1` to `0.31.3`.

Closes #757


xtqqczze:
Since `procps` is as yet unreleased, perhaps it would make sense to [patch](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#the-patch-section) `uucore` to use `main` branch of `coreutils`?

